PR #751, opened 2014-06-19
added a component api

bentedder:
A bit of an experiment. Any component can be accessed/scraped by visiting:
http://rizzo.lonelyplanet.com/component/COMPONENTNAMEHERE?param=value&param=value

So for example:
http://rizzo.lonelyplanet.com/component/tile?title=article%20title&link_url=http://www.ebay.com&image_alt=test

It's a first attempt at this, so comments appreciated, and no rush to merge. But testing locally provides pretty decent results, and lets us scrape UI components into microsites (php).

### DIFF
--- a/app/controllers/layout_controller.rb
+++ b/app/controllers/layout_controller.rb
@@ -13,4 +13,8 @@ class LayoutController < ActionController::Base
     render layout_details[:template], layout: layout_details[:layout], locals: get_layout_config(params[:route])
   end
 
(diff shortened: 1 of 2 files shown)


stevebrownlee:
@bentedder so.... is there any momentum behind this PR, or should we close it?

bentedder:
meh. kill it. seems like bigger/better things are in the works.

PR #884, opened 2014-08-13
the type is meant to be a mime type

chee:
as in rfc 2045
and in this case, it's not even mustache it's just html

i realise this is just a trick but don't let's act like wild animals


chee:
successfully marged ad clothed

PR #1173, opened 2014-12-30
Fixed picker.js throwing random requirejs script error on development

WunderBart:
In Community (development), this script error is rather constant than random: 

screenshot 2014-12-30 15 01 18

Putting `picker` components in `requirejs` config seems to fix the issue. Console is clean :clap:

### DIFF
--- a/app/assets/javascripts/lib/components/datepicker.js
+++ b/app/assets/javascripts/lib/components/datepicker.js
@@ -4,7 +4,7 @@
 //
 // ------------------------------------------------------------------------------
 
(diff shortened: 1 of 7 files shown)


remybach:
This looks good to me :+1: 

Just one note for all @lonelyplanet/ui devs: This is going to need updating of `requirejs.yml` in apps that use Rizzo (just in case you stumble across any issues).

ps: \\o/ glad to finally be rid of this. I seem to recall it featuring in TrackJS a lot.

PR #1178, opened 2015-01-06
Bart fix timeago

WunderBart:
There's a problem in Community where `timeago` randomly simply doesn't init at all (mostly on first visit to random threads). The worst part is that it doesn't leave any error trace whatsoever. Thought it was `require.js`-specific but couldn't find any culprit there.

This is the first (and hopefully final) step to make `timeago` work properly each time.


remybach:
This all looks ok to me... :ship:

PR #1179, opened 2015-01-07
If remove core layout

Ianfeather:
Not for merging yet. Opening early for discussion. 

Hotels and waldorf both still use the `core` layout which basically is a fixed width version of the site scaffold. Now that most of the site is responsive we really want to deprecate any notion of building to a fixed width so this PR removes that layout. 

Fixed width sites in these apps would set an instance variable to toggle that 'mode'. They essentially get the same media query treatment as old IE. As we pick off the last few pages we can remove the variable and, eventually, remove the core_fixed.css file in this PR.

The main benefit to this is that we can remove the `responsive` class from the body as it will no longer act as a toggle. The next PR would remove all mentions of that class in the css, simplifying it a lot.


app/lib/layout_support.rb
```diff
@@ -36,7 +36,6 @@ def layout_options
         default_title:  false
       },
       modern: {
-        responsive:     true,
```
Ianfeather: These were just duplicated in the defaults

app/docs/page-layout/core-layouts.md
```diff
@@ -9,7 +9,6 @@ and the core content is rendered withing .row--content.
 There are three main types of layout currently available using Rizzo, each of which have preview pages:
```
bentedder: now there are 4 main types :)

app/views/layouts/custom/_head.html.haml
```diff
@@ -15,7 +15,7 @@
 
 %link{ rel: "icon", type: "image/gif", href: "https://www.lonelyplanet.com/favicon.ico" }
 
-%meta{ name:"viewport", content: responsive ? "width=device-width, initial-scale=1" : "width=1024" }
+%meta{ name:"viewport", content: responsive && !@fixed_width_layout ? "width=device-width, initial-scale=1" : "width=1024" }
```
bentedder: so, is there a case where responsive would be true, and fixed_width_layout would be true? not sure i understand this bit...

app/docs/page-layout/core-layouts.md
Ianfeather: There are three. Core is deleted :)

app/views/layouts/custom/_head.html.haml
Ianfeather: For requests coming in via from the 'responsive' layout, responsive is set to true. Because fixed width sites will now be using the responsive layout, they'll be forced through this path also so we need this flag to differentiate. We will be able to remove it once we no longer have any fixed width pages. 

This flag has to be here as they don't use the config in layout_support.

Does that make sense?

app/views/layouts/custom/_head.html.haml
```diff
@@ -15,7 +15,7 @@
 
 %link{ rel: "icon", type: "image/gif", href: "https://www.lonelyplanet.com/favicon.ico" }
```
i-like-robots: gif? I don't think the type attribute has any effects but still.

app/views/layouts/custom/_head.html.haml
bentedder: ah. i didn't realize everything was using the responsive layout. makes sense now.

app/views/layouts/custom/_head.html.haml
Ianfeather: Nice spot

bentedder:
looks good to me +1

PR #1180, opened 2015-01-08
change dot to marker.

gcziprusz:
BEFORE:
screen shot 2015-01-08 at 2 34 42 pm
 
AFTER:
screen shot 2015-01-08 at 2 35 04 pm

### DIFF
--- a/app/assets/javascripts/lib/components/poi_list.js
+++ b/app/assets/javascripts/lib/components/poi_list.js
@@ -45,7 +45,7 @@ define([
     }
 
     if (map.marker) {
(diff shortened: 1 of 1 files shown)


stevebrownlee:
QQ: Who requested this change? Are we sure that this change is needed?

stevebrownlee:
:+1:

gcziprusz:
When the map switches from static to dynamic mode it looks like the hotel icon is removed.
This change will make sure that the hotel/center pin is consistent between the static and dynamic modes.

PR #1185, opened 2015-01-12
Fix breaking single apostrophes in tweets

remybach:


app/helpers/social_helper.rb
```diff
@@ -4,7 +4,7 @@ def tweetify(args = {})
     max_length = args[:max_length] || 138
     short_url_length = args[:short_url_length] || 25
 
-    title = args[:title] || ''
+    title = args[:title].sub('\'', '%27') || ''
```
Ianfeather: If `args[:title]` is nil this won't fall back to the empty string it will throw an error. You're gonna need something like:

`title = args[:title].present? ? args[:title].sub('\'', '%27') : ''`

Ianfeather:
+1

PR #1193, opened 2015-01-21
Adding a class to allow a left nav item to appear disabled

stevebrownlee:
> Trello issue: https://trello.com/c/6WkjO9OP/617-link-not-working-properly

## Description
The left nav links on The World view were navigating to another view before the JavaScript loaded.  I added the `.disabled` class so that users had the indication that they could not be clicked on.

## Verification Tasks
Test this change with Waldorf pull request https://github.com/lonelyplanet/waldorf/pull/1794


amccadams:
:+1:

PR #1195, opened 2015-01-22
ability deliver media query styles to old browsers

gcziprusz:
Add an option to output the styles for browsers that do not support media query styles.

For example, we would like to be able to apply the `width` and `display` to IE7/8 in addition to `wide-view` media query block.

```.group--highlight
    +respond-to(wide-view, $fallback: true, $legacy: true)
      width: 50%
      display: block

### DIFF
--- a/app/assets/stylesheets/sass/mixins/_responsive.sass
+++ b/app/assets/stylesheets/sass/mixins/_responsive.sass
@@ -12,7 +12,7 @@
 //
 // [/doc]
 // -----------------------------------------------------------------------------
-@mixin respond-to($screen-width, $fallback: true)
+@mixin respond-to($screen-width, $fallback: true, $legacy-support: false)
 
   // We have to set up a blank variable here or sass errors out
   $breakpoint: ''
@@ -37,6 +37,10 @@
 
   @else
     @media only screen and (min-width : #{$breakpoint}px)
+      @content
+
+  @if ($legacy-support)
+    .ie7 &, .ie8 &
       @content
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by i-like-robots** [comment]
The `$fallback` option will already output the declarations without nesting the entire block within a media query - is this for situations where the global `$is-ie` flag cannot be or is not used?

With this solution to that problem the `$fallback` and `$legacy` arguments can be redundant, as this could be output:

```css
.selector {
  width: 50%;
  display: block;
}

.ie7 .selector,
.ie8 .selector {
  width: 50%;
  display: block;
}
```

I feel like the behaviour of the mixin should probably be controlled at a global level, similar to how the `$is-ie` flag is used to generate the separate IE stylesheets but then having a mixin behave differently between projects would be weird too...

Ianfeather:
Is the issue just that you aren't serving an ie stylesheet? Because the mixin does exactly what you describe. If we wanted to remove the ie stylesheets altogether we could add matt's code but having both seems redundant

gcziprusz:
We opted for the extra param in the existing `+respond-to` mixin with hopes that we can take the `.ie8` part out when we drop support for it without having to update the client code. 
I do like @i-like-robots 's `+legacy-respond-to` idea. We don't at the moment have IE stylesheets and would like the option to apply the wide-screen media query styles to IE7/8 without having to duplicate the styles. 
Thanks for the feedback will do whatever you guys see as the best option going forward.

gcziprusz:
@i-like-robots @Ianfeather do you guys like the new `respond-to-` addition? We could add an IE stylesheet and set globals but seems redundant to duplicate the styles specifically for IE8. We are already using the exact mixin in waldorf so it would good to have in rizzo. 
Thanks

stevebrownlee:
:+1:

PR #1197, opened 2015-01-23
Add external app_core

Ianfeather:


config/requirejs.yml
```diff
@@ -19,6 +19,7 @@ paths:
 
 modules:
   - name: 'infobox'
+  - name: 'app_core_external'
   - name: 'app_core'
```
bentedder: so is there a point to actually bundling app_core as its own module now? does anything actually use it raw?

Ianfeather: Yeah. All the other apps include app_core as an external dependency. e.g. `app_lodgings` will include app_core but not in the bundle. That way we can cache it separately.

bentedder:
@webdeveric has confirmed this works locally. going to merge this in.

PR #1198, opened 2015-01-26
If remove responsive core

Ianfeather:
Removes the .responsive flag from (almost) everything under `/stylesheets/core`.

This basically means secondary nav and footer need checking

### DIFF
--- a/app/assets/stylesheets/core/_grid.sass
+++ b/app/assets/stylesheets/core/_grid.sass
@@ -316,26 +316,22 @@
   box-sizing: border-box
   -moz-box-sizing: border-box
   padding: 5px
+  width: 100%
 
-  // This doesn't really need to hide behind a flag but for
-  // consistency with current Rizzo, it does.
-  .responsive &
-    width: 100%
+  +respond-to(320)
+    width: percentage(1 / 2)
 
-    +respond-to(320)
-      width: percentage(1 / 2)
+  +respond-to(580)
+    width: percentage(1 / 3)
 
-    +respond-to(580)
-      width: percentage(1 / 3)
+  +respond-to(860)
+    width: percentage(1 / 4)
 
-    +respond-to(860)
-      width: percentage(1 / 4)
+  +respond-to(1020)
+    width: percentage(1 / 6)
 
-    +respond-to(1020)
-      width: percentage(1 / 6)
-
-    +respond-to(1120)
-      width: percentage(1 / 8)
+  +respond-to(1120)
+    width: percentage(1 / 8)
 
   > .tile
     min-width: 0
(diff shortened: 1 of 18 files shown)


app/assets/stylesheets/core/_grid.sass
```diff
@@ -316,26 +316,22 @@
   box-sizing: border-box
   -moz-box-sizing: border-box
```
i-like-robots: Autoprefixer should handle all of our prefixing needs

app/assets/stylesheets/core/layouts/modern/_footer.sass
```diff
@@ -11,77 +11,55 @@
...
-  margin: 0 20px 15px 0
-  .responsive &
-    margin: 0
-    +respond-to(wide-view)
-      float: left
-      margin: 0 20px 15px 0
+  border-bottom: 1px solid $input-border
+  +respond-to(wide-view)
+    border: none
+    padding: 0
```
i-like-robots: Not sure there's any padding to override

app/assets/stylesheets/core/layouts/modern/_footer.sass
```diff
@@ -11,77 +11,55 @@
...
+  +respond-to(wide-view)
+    float: right
+    width: 240px
+    padding: 0
+    margin-right: 10px
+
+  @if $is-ie == true
+    width: 200px
 
   .btn--green
```
i-like-robots: we should really author buttons/badges etc with both the object and modifier (`class="btn btn--green"`) so we're not targeting specific visual stuff.

app/assets/stylesheets/core/layouts/modern/_footer.sass
```diff
@@ -93,36 +71,28 @@
...
-    display: none
-    .responsive &
-      display: block
-      +respond-to(wide-view)
-        display: none
+    +respond-to(wide-view)
+      display: none
 
   .accordion__title--footer
     float: right
```
i-like-robots: I think the accordion is only used for the site footer, it's begging for a clean out soon.

app/assets/stylesheets/core/layouts/modern/_footer.sass
Ianfeather: It has the object and modifier. I've added a component specific modifier now and targeted that.

app/assets/stylesheets/core/layouts/modern/_footer.sass
Ianfeather: Totally

Ianfeather: Could be much much simpler

i-like-robots:
:+1:

PR #1203, opened 2015-01-27
Fix flights autocomplete misbehaviour

remybach:


app/assets/javascripts/lib/widgets/flights_autocomplete.js
```diff
@@ -22,7 +22,7 @@ define([
       success: function(data, textStatus, request) {
         var userCurrency, $currencySelect;
 
-        geoIPCountryCode = request.getResponseHeader("X-GeoIP-CountryCode");
+        geoIPCountryCode = request.getResponseHeader("X-GeoIP-CountryCode") || "GB";
```
remybach: fwiw: Deliberately committed this as it's better than there being an error if something goes wrong I reckon.

app/assets/javascripts/lib/widgets/flights_autocomplete.js
```diff
@@ -98,12 +98,12 @@ define([
...
   };
 
-  FlightsWidgetAutocomplete.prototype.onSelectCity = function(el) {
+  FlightsWidgetAutocomplete.prototype.onSelectCity = function(el, event, currentField) {
     var selectedCode = $(el).data("airport").slice(0, -4),
         selectedValue = $(el).text();
 
     selectedValue = selectedValue.substring(0, selectedValue.indexOf(")"));
-    if (this.el === "#js-from-city") {
+    if (currentField === "#js-from-city") {
```
i-like-robots: Is it possible to use `$(el).is("#js-from-city")` instead of passing round the selector?

app/assets/javascripts/lib/widgets/flights_autocomplete.js
i-like-robots: :+1:

app/assets/javascripts/lib/widgets/flights_autocomplete.js
remybach: No, because at this point `$(el)` is referring to the list item in the autocomplete list.

i-like-robots: Ah, then this is probably more elegant than trawling the DOM!

i-like-robots:
Allrighty +1

PR #1205, opened 2015-01-28
Flights widget updates.

remybach:


app/assets/javascripts/lib/widgets/flights_autocomplete.js
```diff
@@ -126,6 +126,23 @@ define([
...
+          if (el == "#js-from-city") {
+            this.$fromAirport.val(airport);
+            this.$fromCity.val(city);
+          } else {
+            this.$toAirport.val(airport);
+            this.$toCity.val(city);
+          }
+        }
+      }.bind(this));
+    }.bind(this));
```
stevebrownlee: Does ie8 support `bind()`?

remybach: Nope, but [we use a polyfill](https://github.com/lonelyplanet/rizzo/blob/master/app/assets/javascripts/app_core.js#L16).

stevebrownlee:
Only one question, and that is irrelevant, then I give :+1:

app/assets/javascripts/lib/widgets/flights_autocomplete.js
stevebrownlee: Wow, I never knew that.  Completely changes how I'm going to write JS here from now on.

PR #1208, opened 2015-01-30
Stop buttons relying on line-height for padding

remybach:


Ianfeather:
Why 1.4em and not just 1? Doesn't this still rely on line height?

remybach:
To some extent, but the `1.4em` is for the actual line height when the text has to go onto two lines, like here:

Ianfeather:
Ok that makes sense. Cool. +1

PR #1209, opened 2015-02-02
fix date format issue with datepicker label

gcziprusz:
Consistently format labels for datepicker.
FROM
datescreen_shot_2015-01-29_at_18 18 59 1

TO
screen shot 2015-02-02 at 3 31 47 pm


Ianfeather:
+1

PR #1212, opened 2015-02-03
Tweaked toggleActive a bit

WunderBart:
- applied [yeoman](https://github.com/lonelyplanet/yeoman/blob/rizzo/javascript-component/templates/component.js) template
- removed listener (no need to use both context & listener at the same time?)
- events are now live within context (support dynamically added toggles - for ex. report form toggle in profile lightbox in Community)

### DIFF
--- a/app/assets/javascripts/lib/components/toggle_active.js
+++ b/app/assets/javascripts/lib/components/toggle_active.js
@@ -1,36 +1,64 @@
-define([ "jquery", "lib/utils/debounce" ], function($, debounce) {
+define([
+  "jquery",
+  "lib/utils/debounce",
+  "polyfills/function_bind"
+], function($, debounce) {
 
   "use strict";
 
-  var LISTENER = "#js-row--content";
+  var defaults = {
+    selector: ".js-toggle-active"
+  };
 
   function ToggleActive(args) {
-    var defaults = {
-      listener: null,
-      context: null,
-      selector: ".js-toggle-active"
-    };
-
     this.config = $.extend({}, defaults, args);
 
-    this.$toggles = $(this.config.selector, this.config.context);
-    this.$context = $(this.config.context || this.config.listener || LISTENER);
+    this.$toggles = $(this.config.selector);
+    this.$context = $(document);
 
-    if (this.$toggles.length) {
-      this._init();
-    }
+    this.init();
   }
 
-  ToggleActive.prototype._init = function() {
+  ToggleActive.prototype.init = function() {
+    this.$toggles.length && this._prepareTogglesAndTargets();
+    this.listen();
+  };
+
+  // -------------------------------------------------------------------------
+  // Subscribe to Events
+  // -------------------------------------------------------------------------
+
+  ToggleActive.prototype.listen = function() {
+    var _this = this;
+
+    this.$context
+      .on("click", this.config.selector, this._handleClick.bind(this))
+      .on(":toggleActive/update", function(e, target) {
+        _this._updateClasses($(target));
+      });
+  };
+
+  // -------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/lib/components/toggle_active.js
```diff
@@ -2,59 +2,44 @@ define([ "jquery", "lib/utils/debounce" ], function($, debounce) {
...
     this.config = $.extend({}, defaults, args);
 
     this.$toggles = $(this.config.selector, this.config.context);
-    this.$context = $(this.config.context || this.config.listener || LISTENER);
+    this.$context = $(this.config.context);
 
-    if (this.$toggles.length) {
-      this._init();
-    }
+    (this.$context.length) && this.init();
```
i-like-robots: No need for parens here, we not calculating the result of an expression

app/assets/javascripts/lib/components/toggle_active.js
```diff
@@ -2,59 +2,44 @@ define([ "jquery", "lib/utils/debounce" ], function($, debounce) {
...
+  // -------------------------------------------------------------------------
+
+  ToggleActive.prototype.listen = function() {
     var _this = this;
 
-    this.$context.on(":toggleActive/update", function(e, target) {
-      _this._updateClasses($(target));
-    });
+    this.$context
+      .on("click", this.config.selector, this._handleClick.bind(this))
```
i-like-robots: We should require the Function.prototype.bind polyfill, just in case.

app/assets/javascripts/lib/components/toggle_active.js
```diff
@@ -63,6 +48,27 @@ define([ "jquery", "lib/utils/debounce" ], function($, debounce) {
...
+  // -------------------------------------------------------------------------
+  // Private Functions
+  // -------------------------------------------------------------------------
+
+  ToggleActive.prototype._prepareTogglesAndTargets = function() {
+    var i, $toggle, $targets;
+
+    this.$toggles.css("cursor", "pointer");
+
+    for (i = 0; i < this.$toggles.length; i++) {
```
i-like-robots: We tend to store the length, it's like 0.00001s faster!

```
for (i = 0, len = this.$toggles.length; i < len; i++) { ... }
```

app/assets/javascripts/lib/components/toggle_active.js
```diff
@@ -81,11 +87,9 @@ define([ "jquery", "lib/utils/debounce" ], function($, debounce) {
 
   ToggleActive.prototype._toggle = function($toggle) {
     this._updateClasses($toggle);
-    this._broadcast($toggle);
+    this.broadcast($toggle);
 
-    if (this.debounce) {
-      this.debounce = null;
-    }
+    if (this.debounce) this.debounce = null;
```
i-like-robots: If this causes any issues with our code style checks - not using braces - we could always do this one liner:

```
this.debounce && (this.debounce = null); 
```

i-like-robots:
Looks good, I'll definitely give it a test in a few apps first though.

app/assets/javascripts/lib/components/toggle_active.js
WunderBart: Hahah, yeah. I removed this one cause,  you know, how many toggles could one actually have? :wink: 
Adding it back! Let's speed this guy up :muscle: :smile:

i-like-robots:
- Mobile primary nav: :-1: 
- Mobile secondary nav: :+1: 
- TTD filters dropdown: :+1: 
- TTD gallery expand: :+1:

i-like-robots:
- Mobile primary nav :+1:

PR #1213, opened 2015-02-04
Fix for `.field__label` & `.field__content` to be actually responsive

WunderBart:
Before these were always 100% wide:
screenshot 2015-02-04 17 07 35

Just had to make shared attributes go first.

### DIFF
--- a/app/assets/stylesheets/core/core_components/_forms.sass
+++ b/app/assets/stylesheets/core/core_components/_forms.sass
@@ -13,25 +13,26 @@
   .input
     float: left
 
+.field__label,
+.field__content
+  @extend %clearfix
+  display: block
+  width: 100%
+  +respond-to(medium-view)
+    float: left
+
 .field__label
   box-sizing: border-box
   padding: 10px 10px 10px 0
-  .field--radio &
-    padding: 0 10px 0 0
   +respond-to(medium-view)
     width: 35%
+
+  .field--radio &
+    padding: 0 10px 0 0
 
 .field__content
   +respond-to(medium-view)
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/core/core_components/_forms.sass
```diff
@@ -13,25 +13,26 @@
   .input
     float: left
 
+.field__label,
+.field__content
+  @extend %clearfix
+  display: inline-block
+  width: 100%
```
i-like-robots: Any point being inline-block if 100% width is forced?

i-like-robots:
Looks fine to me +1

app/assets/stylesheets/core/core_components/_forms.sass
WunderBart: It's shared attr cause they go inline above `medium-view` (width changes to 35% & 65%)

i-like-robots: If there is any whitespace in the markup between the two elements will they still wrap?

WunderBart: It's been working fine all the time - it's just fix after removing all `.responsive` classes

chee:
me too +1

app/assets/stylesheets/core/core_components/_forms.sass
i-like-robots: Ahh, I see they float - my misunderstanding. I'd probably just roll with `display: block` rather than `display: inline-block; width: 100%`.

PR #1214, opened 2015-02-04
This makes the map POI in IE8 look better.

amccadams:
##### This is an IE8 bug fix.  
 The POI descriptions currently have a transparent background and they overlay the map itself.  It looks bad.  This fix adds `respond-to-legacy-support` to the `poi_list.sass` and places the description below the map.

* To check this PR open a hotel page in IE8 and check that the POIs descriptions are below the map.


stevebrownlee:
Looks much better  :+1:

PR #1216, opened 2015-02-05
bowerize deparam

chee:
gets us a step closer to no coffee left in rizzo.

this has belonged in vendor 4eva & 5eva


app/assets/javascripts/lib/page/controller.js
```diff
@@ -3,7 +3,7 @@ define([
   "lib/mixins/events",
   "lib/page/pushstate",
   "lib/mixins/page_state",
-  "lib/utils/deparam"
+  "jquery-bbq-deparam/jquery-deparam"
```
i-like-robots: should this path be prefixed with `vendor/`?

i-like-robots: (Or added to RequireJS config)

chee: no, the vendor/assets are served in the same place as the app/assets

i'm not wrong about this am i?

i'm pretty sure that's right

chee:
@matth  replied

app/assets/javascripts/lib/page/controller.js
i-like-robots: You are right

i-like-robots:
:ship:

PR #1219, opened 2015-02-09
Fix the 'story behind the picture' feature in hero banners

remybach:
Before:

After:
story-behind

### DIFF
--- a/app/assets/stylesheets/components/_banner.sass
+++ b/app/assets/stylesheets/components/_banner.sass
@@ -95,13 +95,22 @@
   line-height: 50px
 
 .hero-banner__image-story
-  +z-layer(top)
-  bottom: -55px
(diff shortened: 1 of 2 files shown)


Ianfeather:
Is the change just to move the text below the X? 

Lots of magic numbers in this PR :/ Mostly the 38px and -45px which stand out. Not sure if you can do anything about them?

remybach:
Yes, but it's a bit deeper than just that... I was initially told that the content would never be more than 250 characters so it was never built to be really flexible. Apparently the content can vary in length (which is what pushed up the "Story behind this picture" text) so that needed to be taken into consideration.

I've found a way around those magic numbers actually (thanks for the nudge).

Ianfeather:
Much nicer +1

PR #1221, opened 2015-02-10
Remove .row-- prefixes from elements that arent in the list

chee:
warning: will break apps

https://github.com/lonelyplanet/waldorf/pull/1808/files

### DIFF
--- a/app/assets/stylesheets/components/_lightbox.sass
+++ b/app/assets/stylesheets/components/_lightbox.sass
@@ -75,7 +75,7 @@ $lightbox-width: 800px !default
     height: 100%
     width: 100%
 
-  .row--billboard
+  .billboard
     display: none
     .content-ready &
       display: block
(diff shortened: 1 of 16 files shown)


chee:
@lonelyplanet/ui should row--faux-banner be left? and row--breadcrumbs? they both are used in a zillion places

Ianfeather:
I'm no longer convinced there's value to this. What do you think? It would be almost easier to just add them to that page showing the page layout. This was proposed as a solution to fix the confusion around how the `rows` in rizzo "work" which I don't feel is much of a problem and certainly hasn't appeared as a problem since. 

This would open up more questions too because changing `.row--image` to `.image` leaves it a bit too unspecific and may cause clashes.

I'm of the opinion it's not worth it, but open to persuasion.

chee:
I am not convinced of it enough to persuade. @lonelyplanet/ui anybody want to persuade?

chee:
closing

PR #1225, opened 2015-02-10
kill error messages

chee:
only required in hotels and not even used
https://github.com/lonelyplanet/hotels/pull/54


remybach:
Lol... :fire: 

:rocket:

PR #1226, opened 2015-02-10
Fix the image stretchy by using percentage 4 width

chee:
https://trello.com/c/R9iWiP01/603-images-are-still-being-stretched-at-larger-screen-sizes-in-the-image-gallery

### DIFF
--- a/app/assets/javascripts/lib/components/gallery.js
+++ b/app/assets/javascripts/lib/components/gallery.js
@@ -84,7 +84,7 @@ define([
   Gallery.prototype._handleEvents = function() {
 
     this.$listener.on(":slider/slideChanged", function() {
-      window.setTimeout(this._afterNavigation.bind(this), +this.slider.config.transition);
+      window.setTimeout(this._afterNavigation.bind(this), this.slider.config.transition * 1.5);
     }.bind(this));
 
     this.$gallery.on("click", ".is-previous", function() {
--- a/app/assets/stylesheets/sass/mixins/_responsive.sass
+++ b/app/assets/stylesheets/sass/mixins/_responsive.sass
@@ -6,7 +6,8 @@
 // Always use this mixin to apply breakpoints so we can support legacy browsers.
 //
 // You can pass in named arguments or integers.
-// Named argument options are: narrow-view (360px), medium-view (600px) and wide-view(1025px).
+// Named argument options are: narrow-view (360px), medium-view (600px),
+// wide-view(980) larger-than-ipad(1025px) and cinema-view(1100).
 //
 // Example of use: http://sassmeister.com/gist/8047633.
 //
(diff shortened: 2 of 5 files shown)


Ianfeather:
This all looks good code-wise. Does the fix the stretchy image? And do you need to remove the `* 0.8` now on padding-top or does that tie in with the 80%?

chee:
the .8 seems to be fine and it gets rid of the stretchy image. feel free to merge

app/assets/javascripts/lib/components/gallery.js
```diff
@@ -84,7 +84,7 @@ define([
   Gallery.prototype._handleEvents = function() {
 
     this.$listener.on(":slider/slideChanged", function() {
-      window.setTimeout(this._afterNavigation.bind(this), +this.slider.config.transition);
+      window.setTimeout(this._afterNavigation.bind(this), this.slider.config.transition * 1.5);
```
remybach: Why `* 1.5`? Shouldn't the `config.transition` that's passed in be updated rather?

app/assets/stylesheets/sass/mixins/_responsive.sass
```diff
@@ -6,7 +6,8 @@
 // Always use this mixin to apply breakpoints so we can support legacy browsers.
 //
 // You can pass in named arguments or integers.
-// Named argument options are: narrow-view (360px), medium-view (600px) and wide-view(1025px).
+// Named argument options are: narrow-view (360px), medium-view (600px),
+// wide-view(980) larger-than-ipad(1025px) and cinema-view(1100).
```
remybach: ron-burgundy-well-that-escalated-quickly

app/assets/javascripts/lib/components/gallery.js
chee: this is just to slightly dejankify the image gallery. for some reason the `_afterNavigation` happens just before the transition is finished

app/assets/stylesheets/sass/mixins/_responsive.sass
chee: i like `larger-than-ipad` and `cinema-view` a lot
we should name them all after devices, it is easier to think about than things like `wide-view` and `large-view`

app/assets/javascripts/lib/components/gallery.js
remybach: I was going to suggest adding a fixed number of ms instead, but either way it feels a bit like grasping at straws and the transition should never be *that* large that passing in a large transition would cause an unexpectedly large gap (if that makes any sense).

app/assets/stylesheets/sass/mixins/_responsive.sass
remybach: Where's the `ultra4k-view` though? :stuck_out_tongue:

app/assets/javascripts/lib/components/gallery.js
chee: the transition length is also set by the number that is passed in

remybach:
Yeah, looking good to me :+1:

app/assets/javascripts/lib/components/gallery.js
chee: it's not in the CSS anymore :O

PR #1229, opened 2015-02-12
Fix twttr issue and travel insurance z-index

remybach:
Thanks to @chee for the clever `onload` solution.


Ianfeather:
Looks fine +1

PR #1232, opened 2015-02-13
Wrap deparam in require

Ianfeather:
We're not going to be updating deparam, wrapShim on requirejs works only intermittently, this requires no config for any other apps. Any objections?


remybach:
Perhaps just a big `NB` comment on top of the file saying that? That way it'll at least be immediately visible in the diff should someone try to update it one day (for any reason).

Ianfeather:
Good call

remybach:
Fine by me :+1:

PR #1233, opened 2015-02-13
Added autofocus & autocomplete attributes to text input

WunderBart:
We need these in Community.
Someone else might need em too. 
Good to have em :]


remybach:
Looks good to me :+1:

PR #1234, opened 2015-02-13
Bt validations

bentedder:
Ran rizzo through some HTML validation and did a tad of clean-up. Would love some :eyes: on it. Especially anybody dealing w/mustache templates, if that will screw with you at all. Also an unnecessary charset on a script tag surrounding the DFP stuff.


bentedder:
and fixed a dumb indentation mistake that I probably made when previewing external routes when using rizzo as a service (/layouts/modern, etc.)

Ianfeather:
+1 Looks fine

PR #1236, opened 2015-02-17
fix maps not expanding on desktop

chee:


chee:
this is no longer needed, the mobile fix fixes desktop too

PR #1244, opened 2015-02-18
allow unescaped HTML in hero-banner content

dhodges:
cukes
```
20 scenarios (20 passed)
61 steps (61 passed)
0m1.213s
```

specs
```
Finished in 0.74282 seconds
168 examples, 0 failures
```

### DIFF
--- a/app/views/components/_hero_banner.html.haml
+++ b/app/views/components/_hero_banner.html.haml
@@ -13,7 +13,7 @@
         %h4.copy--h4
           Story behind the picture
         %p.copy--body
-          = properties[:image_caption]
+          != properties[:image_caption]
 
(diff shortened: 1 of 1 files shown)


chee:
in private i voiced concern that this would be the only value unescaped, and so it's something extra to remember, but on the whole i have no real misgivings with this pull request and welcome it gladly with open arms

chee:
+1

dhodges:
the degree of change involved here is not to be taken lightly - specifically, this new ability to pull raw HTML into a component template. But generally we have to consider that if the editors have asked for this feature, then they are fully alive to the awe and majesty of their newly awakened powers.

PR #1231, opened 2015-02-13
Amends from Ruby `2.2.0` environment

marckysharky:

PR #1246, opened 2015-02-19
remove duplicate :date attribute from form inputs

robuye:
fixes errors:
```
/app/views/components/forms/_radio_buttons_field.html.haml:4: warning: duplicated key at line 4 ignored: :data
/app/views/components/forms/_text_area.html.haml:2: warning: duplicated key at line 2 ignored: :data
```

these _warnings_ started showing in Rails 4.2.

I'm not sure anyone uses these features or where we could test the behaviour is correct? tests are passing...

@remybach can you have a look? :pray:

### DIFF
--- a/app/views/components/forms/_radio_buttons_field.html.haml
+++ b/app/views/components/forms/_radio_buttons_field.html.haml
@@ -1,7 +1,7 @@
 = render :layout => 'components/forms/partials/field', :locals => { id: properties[:id], label: properties[:label], extra_class: "field--radio #{properties[:extra_class]}", tooltip: properties[:tooltip] } do
   - properties[:options].each do |option|
     %span.input__container.js-input{ class: properties[:error] ? 'field__input--error' : nil }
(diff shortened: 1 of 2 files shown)


remybach:
I thought @marckysharky had already fixed this not too long ago :confused:

remybach:
Ah, yes... in lonelyplanet/rizzo#1231 which hasn't been merged yet.

From a ruby perspective I'm not sure which solution is better... I'll let you two fight it out.

robuye:
ah, I didn't notice it. Also our approaches result in slightly different output. I do merge both data attributes and we may end up with `data-rules` being empty.

For Sharky it's either rules or data, never both together. So his p/r is more backward compatible, mine handles things slightly different way.

robuye:
I couldn't find any existing references that would rely on rules for radio buttons or checkboxes so backward compatibility (as in Sharky's p/r) is not really something to worry about.

_merging away_

PR #1247, opened 2015-02-23
fix last-input in ff, & correct behaviour

chee:


chee:
bep

app/assets/javascripts/lib/utils/last_input_device.js
```diff
@@ -23,13 +20,13 @@ define(function() {
...
+
+    if (userDevices[deviceType]) return;
 
     // announce the presence of each device as it's used for the first time
-    if (!listener) { return; }
-    event = document.createEvent("CustomEvent");
-    event.initCustomEvent(":device/" + deviceType);
-    listener.dispatchEvent(event);
+    userDevices[deviceType] = true;
+    listener.dispatchEvent(new CustomEvent(":device/" + deviceType));
```
jcreamer898: Doesn't look like `CustomEvent` is supported in IE8? https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent

app/assets/javascripts/lib/utils/last_input_device.js
```diff
@@ -23,13 +20,13 @@ define(function() {
       document.documentElement.className += " last-input-" + deviceType;
     }
 
-    userDevices[deviceType] = true;
+    if (!listener) return;
+
+    if (userDevices[deviceType]) return;
```
jcreamer898: Could combine these 2 if statements... 

```js
if (!listener || userDevices[deviceType]) return;
```

app/assets/javascripts/lib/utils/last_input_device.js
jcreamer898: Never mind, realizing this code only affects "modern" browsers.

remybach:
:+1:

PR #1248, opened 2015-02-23
anchor target

chee:
for the styleguide so you can link to specific colours and icons. any whatever else might have a text filter some day

### DIFF
--- a/app/assets/javascripts/lib/styleguide/svg.js
+++ b/app/assets/javascripts/lib/styleguide/svg.js
@@ -48,4 +48,8 @@ define([ "jquery" ], function($) {
     showMatchingIcons(this.value);
   });
 
+  if ($searchBox.val()) {
+    showMatchingIcons($searchBox.val());
+  }
+
 });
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/lib/styleguide/svg.js
```diff
@@ -10,18 +10,18 @@ define([ "jquery" ], function($) {
...
-      $intro.removeClass("is-closed");
-    }
-  };
+      showMatchingIcons = function(query) {
+        $iconCards.addClass("is-hidden").each(function() {
+          var element = $(this);
+          element.data("icon").match(query) && element.removeClass("is-hidden");
+        });
+
+        if ($iconCards.filter(".is-hidden").length) {
```
jcreamer898: This `if` statement can actually be simplified with `toggleClass`...

```js
$intro.toggleClass("is-closed", $iconCards.filter(".is-hidden").length);
```

http://api.jquery.com/toggleclass/#toggleClass-className-state

chee: do you think that is more clear?

remybach: TIL!

I think the `toggleClass` is neater, but I'd change it to `.length > 0` so that it's clearer the second argument is boolean.

remybach:
:rocket:

PR #1249, opened 2015-02-24
Make toggle active initialize a bit more synchronously and add "listening" event.

remybach:


app/assets/javascripts/lib/components/toggle_active.js
```diff
@@ -20,8 +20,7 @@ define([
   }
 
   ToggleActive.prototype.init = function() {
-    this.$toggles.length && this._prepareTogglesAndTargets();
-    this.listen();
+    this.$toggles.length && this._prepareTogglesAndTargets() && this.listen();
```
i-like-robots: Maybe instead of requiring `_prepareTogglesAndTargets` to return something truthy you could do something like:

```
this.$toggles.length && (this._prepareTogglesAndTargets(), this.listen());
```

Bit dirty but one line!

i-like-robots:
squerge

PR #1252, opened 2015-02-25
added tracking to search autocomplete

bentedder:
adding in tracking for autocomplete search results.

### DIFF
--- a/app/assets/javascripts/lib/core/nav_search.js
+++ b/app/assets/javascripts/lib/core/nav_search.js
@@ -54,12 +54,12 @@ define([ "jquery", "autocomplete" ], function($, Autocomplete) {
 
   NavSearch.prototype.onItem = function(el) {
     var $el = $(el);
-    window.location = $el.attr("href");
     window.lp.analytics.api.trackEvent({
       category: "search",
       action: "autocomplete",
       value: $el.attr("href")
     });
+    window.location = $el.attr("href");
   };
 
   $(document).ready(function() {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/lib/core/nav_search.js
```diff
@@ -53,7 +53,13 @@ define([ "jquery", "autocomplete" ], function($, Autocomplete) {
   }
 
   NavSearch.prototype.onItem = function(el) {
-    window.location = $(el).attr("href");
+    var $el = $(el);
+    window.location = $el.attr("href");
+    window.lp.analytics.api.trackEvent({
```
jcreamer898: Not sure, but you MIGHT want to do the `window.location` part after the track event. Could run into a race condition. In theory, the page will change the moment you set a location.

bentedder: good catch. will change.

app/assets/javascripts/lib/core/nav_search.js
```diff
@@ -53,7 +53,13 @@ define([ "jquery", "autocomplete" ], function($, Autocomplete) {
   }
 
   NavSearch.prototype.onItem = function(el) {
-    window.location = $(el).attr("href");
+    var $el = $(el);
```
jcreamer898: Another, teeny tiny thing, but you could also store `$el.attr("href");` as a variable, so you're not calling `attr` twice.

```js
var $el = $(el),
      href = $el.attr("href");
```

remybach: Sorry to butt in, but I think just using `el.href` would be the quickest here.

jcreamer898:
:ship:

PR #1254, opened 2015-02-26
de-shrinkwrap

bentedder:


chee:
kill it dead,\.

hi @Ianfeather

bentedder:
for reference: shrinkwrapping sucks in that it pulls in optionaldependencies by default. failing builds constantly, so stripping it for now.

PR #1255, opened 2015-02-27
rfc 2045 reborn

chee:
one small step for sed, a normal person foot sized step for rfc2045 being satisfied.
as it was being broken for no good reason.


remybach:
:+1:

PR #1256, opened 2015-02-27
Serializeform.js, all them tests is sure passing

chee:


app/assets/javascripts/lib/utils/serialize_form.js
```diff
@@ -0,0 +1,71 @@
+define([
+  "jquery"
+], function($) {
```
remybach: Can this not go on one line and then expand later *if* any extras are required?

chee: i suppose, but this way it's more obvious what changed in the diff when something new is added. :o either way is fine with me. YOU WANT ME TO CHANGE IT?

remybach: Meh... not a biggie at all.

remybach:
:rocket:

PR #1259, opened 2015-03-02
Fix issue with events  not working

jcreamer898:
The container div has been places absolutely on the placeholder. An `mv--inline` class has been put on the container as of January 25th, this has caused events not fire on the placeholder.

https://github.com/lonelyplanet/rizzo/blame/5c329b314761290f7b61288e02e084e795c1023b/app/assets/stylesheets/core/utilities/_responsive.sass#L114

Simply changing where the events are bound to the container seems to resolve the issue.


gcziprusz:
:ship:  it

PR #1264, opened 2015-03-09
Remove contentious Israel and Jerusalem icon

patientfrenzy:
We have had a request to remove the icon used for Israel and Jerusalem - this does it.

```
528 specs in 6.772s.
>> 0 failures
```


beatrichartz:
:+1:

PR #1265, opened 2015-03-09
Maps were being loaded any click or mouseover

jcreamer898:
Please check by,

1. Loading up a map on somewhere like hotels
2. Move mouse or click in static map
3. Ensure that map is created correctly with all POIs
4. Ensure subsequent clicks do not attempt to continually recreate the map


JoeShep:
:+1: Map loads once.

PR #1267, opened 2015-03-09
Ds add snapto on range slider

davidthesavage:
This PR adds the option of `snapDateAt` to the range_slider component, which is specified on the data property of the haml template as `snap_date_at`. 

If this option is provided, the range slider will snap to 24 unit steps rather than allowing the granularity of 1 unit at the value specified in `snap_date_at`.

The reasoning for this is to fix an issue where filters were allowing hour granularity, even though the labels for the filter were showing at the day level of granularity. This caused the UI to send values that would filter out TTD that would appear to the user to be within the selected range.


stevebrownlee:
:+1: Make sure you squabaserge

PR #1269, opened 2015-03-16
Bart update autocomplete styles

WunderBart:
primary_nav_search
page_hopper
screenshot 2015-03-16 13 54 37
screenshot 2015-03-16 13 54 57

### DIFF
--- a/app/assets/stylesheets/components/_autocomplete.sass
+++ b/app/assets/stylesheets/components/_autocomplete.sass
@@ -5,6 +5,16 @@
 // -----------------------------------------------------------------------------
 
 .autocomplete
+  position: relative
+  input
+    background-image: url("#{image_path('shared/spinner.gif')}")
+    background-repeat: no-repeat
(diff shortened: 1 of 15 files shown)


app/assets/stylesheets/components/_autocomplete.sass
```diff
@@ -5,6 +5,17 @@
 // -----------------------------------------------------------------------------
 
 .autocomplete
+  position: relative
+  input
+    float: none !important
```
stevebrownlee: Is the `!important` necessary here?

stevebrownlee:
Just one question. If the answer is yes, then :+1:

app/assets/stylesheets/components/_autocomplete.sass
WunderBart: It's not! :D It was actually needed for flights widget. I will fix it there instead. Good catch +1

PR #1271, opened 2015-03-19
Fix inconsistent button height bug

WunderBart:
`.btn--outline` is 2px higher than any other button.

Before:
screenshot 2015-03-19 01 26 50
After:
screenshot 2015-03-19 01 27 05


stevebrownlee:
Epic change  :+1:

PR #1272, opened 2015-03-23
Updating style guide for es2015 rules

jcreamer898:
Adding in rules for working with ES2015. This is totally up for any collaboration.

### DIFF
--- a/app/docs/js/code-style.md
+++ b/app/docs/js/code-style.md
@@ -25,12 +25,12 @@ Disagree with any of the settings? Make a PR and open a discussion.
 
 * Name collections (arrays, objects, sets, maps) in plural, ie: `badger` is a thing, `badgers` is a collection of `badger`s
 
-* Please preface functions that are intended to be private with an underscore:
+* Please preface functions that are intended to be private (I.E., not intended to be called) with an underscore:
 
 ```javascript
 // private function convention
 var _foo = function() {
-  // statements 
+  // statements
 }
 ```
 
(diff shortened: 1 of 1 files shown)


davidthesavage:
:+1:

app/docs/js/code-style.md
```diff
@@ -147,4 +124,155 @@ $(selector)
...
+], function(Godliness, Cleanliness) {
+```
+
+## ES6 (2015) Guide
+
+### Variable Declarations
+
+* Use `let` almost always, and declare in correct scope. No more need to hoist all declarations
+
+* Use `const` when needing a **primitive** value (aka string, number, boolean) to never change
```
stevebrownlee: Why only primitives? I can definitely see the case where you want a constant object declaration that can be modified, but not be redefined with another object.

jcreamer898: Will modify, and mention some things about it. Thanks.

app/docs/js/code-style.md
```diff
@@ -147,4 +124,155 @@ $(selector)
...
+let name = this.user.name,
+    age = this.user.age,
+    location = this.user.location;
+
+// good
+let { name, age, location } = this.user;
+```
+
+### Arrow functions
+* Use arrow functions possible
```
stevebrownlee: Option 1: "Use arrow functions **whenever** possible"

Option 2: "Arrow functions should be used by default, except in rare cases where it's not possible"

jcreamer898: :+1:

stevebrownlee:
I would also like you to add a section about using Array Comprehensions whenever simple transformations or filtering needs to take place on an array.

jcreamer898:
@stevebrownlee you mean like `[for (i of numbers) i*i ]` or `map`, `filter`, etc?

stevebrownlee:
for...of

stevebrownlee:
Squerge :+1:

PR #1273, opened 2015-03-23
added @no_nav check for apps that don't need nav bar

bentedder:
This PR allows apps to not use the blue bar (specifically for some BizDev projects).

**Example implementation:**
```
layout "responsive"

def index
    @no_nav = true
    render "mypage"
end
```


gcziprusz:
:shipit:

PR #1274, opened 2015-03-24
Remove country selection based on user currency.

remybach:
https://trello.com/c/gSe42kdj/337-remove-autofill-of-home-country-for-car-rental


stevebrownlee:
Looks good :+1:

PR #1279, opened 2015-04-02
Adding a possibility to create links with set target in nav items

nidalaa:
@WunderBart what do you think?

### DIFF
--- a/app/views/components/navigation/_nav_item.html.haml
+++ b/app/views/components/navigation/_nav_item.html.haml
@@ -14,12 +14,12 @@
 
 - if item[:slug]
 
-  %a.nav__item.js-nav-item{ data: item[:analytics], class: [item_classes, active, item[:extra_class]], href: item[:slug], itemprop: "url", aria: { label: active.present? ? "Current page" : nil } }<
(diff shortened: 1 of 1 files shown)


WunderBart:
Could you present some use case for this?

stevebrownlee:
@nidalaa What is the purpose of this change? Can you provide a brief description of what you are trying to accomplish? Thanks.

nidalaa:
@stevebrownlee  The main change is adding `target: item[:target]`. Thanks to this when I am creating a list of nav items, I can set target to `_blank`, so the link will be open in new tab. This is a main reason of this p/r(rest changes are just about attribute order).

stevebrownlee:
Well that sounds like a great addition.  How do I test it?

nidalaa:
@stevebrownlee  I think that the easiest way is to add:
`target: "_blank"` to secondary items in style guide application(https://github.com/lonelyplanet/rizzo/blob/master/app/lib/rizzo_app.rb#L35).

stevebrownlee:
Works as described. :+1:

PR #1280, opened 2015-04-03
Adding new ad unit size for bottom of destinations DFP ad

stevebrownlee:
Per Tim Daugherty, the ad unit at bottom on destinations needs to be 728x90 on desktop and 300x250 on mobile.  This adds the new size.


gcziprusz:
:shipit:

PR #1282, opened 2015-04-08
Add in styleguide, and core for destinations pages

jcreamer898:
This simply adds a new folder to rizzo for destinations next, no existing code is modified. Also an npmignore added.


davidthesavage:
:+1:

PR #1284, opened 2015-04-08
Fix issue with jscs, add submodule update to npm

jcreamer898:
Update fonts, create pre-install script to grab them


stevebrownlee:
:+1:

PR #1290, opened 2015-04-10
added no_footer var for external sites

bentedder:
Very similar to #1273, just removes the footer when not needed.


JoeShep:
:+1:

PR #1295, opened 2015-04-17
Fix font urls to be case correct

davidthesavage:


davidthesavage:
:+1: :+1: from @jcreamer898

PR #1298, opened 2015-04-28
Fix no-3d-support

webdeveric:
@stevebrownlee Can someone on your team take a look at this one?

I am seeing `no-3d-support` on the html element which is wrong since `translate3d()` has been available in Firefox for a while.

I've updated feature_detect.js to address this issue.

### DIFF
--- a/app/assets/javascripts/lib/core/feature_detect.js
+++ b/app/assets/javascripts/lib/core/feature_detect.js
@@ -101,10 +101,10 @@ define([ "jquery" ], function($) {
   features.transform = function() {
     var
       transforms = {
+        transform: "transform",
         webkitTransform: "-webkit-transform",
         MozTransform: "-moz-transform",
-        msTransform: "-ms-transform",
-        transform: "transform"
+        msTransform: "-ms-transform"
       },
       transform;
 
--- a/app/assets/javascripts/lib/core/feature_detect.js
+++ b/app/assets/javascripts/lib/core/feature_detect.js
@@ -42,7 +42,7 @@ define([ "jquery" ], function($) {
     document.body.insertBefore(el, document.body.firstChild);
 
     if (transform) {
-      el.style[transform.css] = "translate3d(1px,1px,1px)";
+      el.style[transform.js] = "translate3d(1px,1px,1px)";
       has3d = window.getComputedStyle(el).getPropertyValue(transform.css);
     }
 


stevebrownlee:
Is Firefox the only browser that was holding this back?

webdeveric:
@stevebrownlee It looks like it. I just did a quick check in Chrome, Safari, IE10/11 and they all had supports-3d where Firefox did not.

stevebrownlee:
Then it might have been in there to continue support for IE 8 and 9... possibly?

webdeveric:
@stevebrownlee IE8/9 do not support `translate3d` and they correctly have the `no-3d-support` class on the `html` element. Browsers that do support `translate3d` should have the `supports-3d` class on the `html` element.

The existing code checks the vendor prefixed transform properties first, then checks the non-prefixed version, which seems backwards to me.

If the browser supports the standards compliant version, it will show up in the `CSS2Properties` object in addition to the vendor prefixed version (`MozTransform` in this case). And if you're in browser that doesn't support transform at all (Old IE), it will not exist in the `CSS2Properties` object.

The issue comes down to assigning `translate3d` to the wrong style property.
It should be using the JS camel case version, not the CSS dashed version of the property name.

Here is a quick example to describe the issue:

**Current behavior in master**

```js
document.documentElement.style["-moz-transform"] = "translate3d(1px, 1px, 1px)";
window.getComputedStyle( document.documentElement ).getPropertyValue("-moz-transform");
```

Returns `none`

**Behavior in my branch**

```js
document.documentElement.style["transform"] = "translate3d(1px, 1px, 1px)";
window.getComputedStyle( document.documentElement ).getPropertyValue("transform");
```

Returns `matrix3d(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 1)`

```js
document.documentElement.style["MozTransform"] = "translate3d(1px, 1px, 1px)";
window.getComputedStyle( document.documentElement ).getPropertyValue("-moz-transform");
```

Returns `matrix3d(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 1)`

In my branch, I use `transform.js` to set the transform property, which results in the correct value being returned from `window.getComputedStyle`.

stevebrownlee:
Now that's probably the best comment I've seen in months  :+1:

PR #1299, opened 2015-04-29
Support for articles widget

davidthesavage:
This adds necessary mixins and classes for articles widget

ALSO fixes the need for a window.load event on text clamping!


stevebrownlee:
:+1:

PR #1308, opened 2015-05-13
Update mixins and remove some spacing where it didn't belong

davidthesavage:
Heading-2 now is no longer a mixin, but changes it's properties at the same breakpoints regardless of it's component. 

Many other mixins updated to be more responsive and accurate to destinations next PSD


gcziprusz:
:+1:

PR #1310, opened 2015-05-21
Support Apple campaign ad sizes

stevebrownlee:
The new sizes are to support the Apple campaign headed our way.  The [1,1] sizes will allow the AdOps team to not show the MPU unit and bottom unit in Destinations and Homepage.


jcreamer898:
:+1:

PR #1311, opened 2015-05-22
Updating the new ad size for the Apple campaign from 970x300 to 1000x300

stevebrownlee:
Title says it all


amccadams:
:+1: That looks like a new ad size to me.

PR #1312, opened 2015-05-25
Fix google breadcrumbs

robuye:
Remove old breadcrumbs as these were invalid and replace with new ones.

For reference:
https://developers.google.com/structured-data/breadcrumbs

Example output:

```
<div class="google-breadcrumbs is-hidden">
  <div id="google-breadcrumb-0" itemprop="child" itemref="google-breadcrumb-1" itemscope="" itemtype="http://data-vocabulary.org/Breadcrumb">
    <a href="http://www.lonelyplanet.com/thorntree" itemprop="url">
      <span itemprop="title">
        Thorn Tree
      </span>
    </a>
  </div>
  <div id="google-breadcrumb-1" itemprop="child" itemscope="" itemtype="http://data-vocabulary.org/Breadcrumb">
    <a href="http://www.lonelyplanet.com//thorntree/categories/country-forums" itemprop="url">
      <span itemprop="title">
        Country forums
      </span>
    </a>
  </div>
</div>
```

passes validation and produces valid output using:
https://developers.google.com/structured-data/testing-tool/

Does not include "Home" breadcrumb, it doesn't make much sense because
site name is going to be used by google anyways.

Does not include very last item because this is going to be a page
title and is not really a proper breadcrumb.

The final output for above example would be:
`www.lonelyplanet.com > Thorn Tree > Country forums`

### DIFF
--- a/app/assets/javascripts/app_core.js
+++ b/app/assets/javascripts/app_core.js
@@ -6,6 +6,8 @@ define([
   "sCode",
   "trackjs",
   "polyfills/xdr",
+  "polyfills/function_bind",
+  "polyfills/array_index_of",
   "lib/page/swipe",
   "lib/core/nav_search",
   "lib/page/scroll_perf",
@@ -13,7 +15,6 @@ define([
   "lib/core/shopping_cart",
   "lib/core/feature_detect",
   "lib/core/place_title_nav",
-  "polyfills/function_bind",
   "lib/core/cookie_compliance",
   "lib/components/toggle_active",
   "lib/components/select_group_manager"
--- a/app/views/components/navigation/_breadcrumbs.html.haml
+++ b/app/views/components/navigation/_breadcrumbs.html.haml
@@ -1,17 +1,26 @@
 - icon_class = properties[:icon] ? "breadcrumbs--icon icon--#{properties[:icon]}--before" : ""
 
-.nav--inline{ itemscope: true, itemtype: "http://data-vocabulary.org/Breadcrumb", class: icon_class }
+.nav--inline{ class: icon_class }
 
   - if properties[:show_home]
-    %a.nav__item.nav__item--breadcrumbs--home.icon--home{href: '/', itemprop:"url"}> Home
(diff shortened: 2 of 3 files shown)


chortlehoort:
I might be afraid of the answer, but is there a way to test this locally?  Also, can you provide a URL where we can see where it's currently broken?  Thanks.

robuye:
thanks for looking into it @chortlehoort !

If you have an app locally that uses rizzo and shows breadcrumbs you can simply switch to this branch and check the difference in the sources (bottom of the page usually). 

See this url from production: http://www.lonelyplanet.com/germany/berlin

The breadcrumb from HTML source is:

```
<div class="nav--inline" itemscope="" itemtype="http://data-vocabulary.org/Breadcrumb"><a class="nav__item nav__item--breadcrumbs--home icon--home" href="/" itemprop="url">Home</a><a class="nav__item js-nav-item nav__item--breadcrumbs" href="http://www.lonelyplanet.com/europe" itemprop="url">Europe</a><a class="nav__item js-nav-item nav__item--breadcrumbs" href="http://www.lonelyplanet.com/central-europe" itemprop="url">Central Europe</a><a class="nav__item js-nav-item nav__item--breadcrumbs" href="http://www.lonelyplanet.com/germany" itemprop="url">Germany</a><a class="nav__item js-nav-item nav__item--breadcrumbs" href="http://www.lonelyplanet.com/germany/berlin-brandenburg" itemprop="url">Berlin &amp; Brandenburg</a><span class="current js-nav-item nav__item nav__item--breadcrumbs">Berlin</span></div>
```

you can copy/paste this to google testing tool https://developers.google.com/structured-data/testing-tool/ to see the errors. It's relatively easy to fix this single error (missing title), but the resulting breadcrumb still won't have proper structure, to see how it should really look you can load sample data in testing page or copy paste code from the p/r description.

robuye:
I'm going to switch Thorn Tree to this branch so the effect will be easier to see.

robuye:
It's live on TT, google already shows proper breadcrumbs for some sites, see first 2 results in screenshot below.

image

@chortlehoort good to go?

PR #1315, opened 2015-05-28
More refined enhancements for Apple campaign

stevebrownlee:
Remove the 1001x300 size for the leaderboard slot since that's not what we're targeting.  Also making the MPU ad on the side rail have the same bgcolor as the body so that when it is set to 1x1, it's not obtrusive by having a different background color.


jcreamer898:
Looks good.

PR #1318, opened 2015-06-03
Adding in the Criteo JavaScript sent to us by their integration team

stevebrownlee:
## Verification steps

1. Point Waldorf to local Rizzo
2. Bring up any page
3. Verify that a request is sent to the Criteo platform (see picture below)

screen shot 2015-06-03 at 10 43 02 am


jcreamer898:
Seems legit...

PR #1321, opened 2015-06-11
Upload new hostelworld logo.

Keqi:


stevebrownlee:
:+1: There is currently a failure in the Rizzo builds that we need to figure, but go ahead and merge

PR #1323, opened 2015-06-18
adds necessary items for hotels datepicker

amccadams:
*Acceptance criteria*
* The user can select check-in and check-out dates and the number of guests.
* When the user clicks on Search we take him to the existing hotels search result page.
* There is a background image using 100% of the width.
* There is a title and subtitle.

Look at bane and Destinations Next PR:
https://github.com/lonelyplanet/bane/pull/92
https://github.com/lonelyplanet/destinations-next/pull/187


app/assets/stylesheets/rizzo_next/_color_theme.scss
```diff
@@ -0,0 +1,5 @@
+// -----------------------------------------------------------------------------
+// States
+// -----------------------------------------------------------------------------
+
+$active: $teal !default;
```
davidthesavage: This file seems unnecessary (and does not look like it is even being included anywhere). I'm also noticing `$active` is defined in the `_color_variables.scss`. This seems it could cause confusion as to which place should be changed if this *is* used, so let's get rid of it if possible.

app/assets/stylesheets/rizzo_next/_color_variables.scss
```diff
@@ -31,3 +31,5 @@ $articles-blurb-color            : #7e838b !default;
 $footer-bottom-link-color        : #959aa0 !default;
 $footer-copyright-color          : #bfc2c6 !default;
 $footer-bottom-icon-border-color : #cccdce !default;
+$active                          : #000080 !default;
+$linkblue                        : #288ad6 !default;
```
davidthesavage: Were `$active` and `$linkblue` carried over from old rizzo code, or is this new colors from the PSD?

app/assets/stylesheets/rizzo_next/_mixins.scss
```diff
@@ -139,3 +139,124 @@ $button-divider-line-width: 124px;
...
+    }
+  }
+}
+
+// -----------------------------------------------------------------------------
+// Size
+// -----------------------------------------------------------------------------
+
+
+@mixin size($size) {
```
davidthesavage: Not liking this mixin, but I'm assuming this is brought in because it was used in the old rizzo datepicker scss. Maybe we can refactor that code to remove this, it just seems to add another layer of confusion.

app/assets/stylesheets/rizzo_next/_color_theme.scss
amccadams: file removed

app/assets/stylesheets/rizzo_next/_color_variables.scss
amccadams: They were carried over from the datepicker in rizzo.  We don't have a PSD mockup for that.

davidthesavage: Makes sense, these are probably fine then.

app/assets/stylesheets/rizzo_next/_color_theme.scss
davidthesavage: Looks like it's still in the PR. You probably need to locally do: `git rm app/assets/stylesheets/rizzo_next/_color_theme.scss` then commit and push

PR #1324, opened 2015-06-18
Updated paths in .npmignore

webdeveric:
@jcreamer898 This file was causing the `app/assets/javascripts/lib` folder to not be included when running  `npm install lonelyplanet/rizzo`. I've updated the paths so that they match folders in the root of rizzo.


webdeveric:
Thumbs up from Creamer.

PR #1327, opened 2015-06-24
Adding target script that enables Criteo to receive the correct DFP tag values

stevebrownlee:
Title says it all. Script comes directly from Criteo integration guide.

## Verification

1. Visit http://assets.lonelyplanet.com/YXBwbGVfdGVzdA==.html
2. Verify that page loads correctly
3. Verify that there is a call to `rta.js`


bentedder:
Verification steps worked. :+1:

PR #1328, opened 2015-06-24
add logs to autocomplete

bentedder:
Returning autocomplete results to flamsteed when user successfully selects an item in the autocomplete.


stevebrownlee:
:+1:

PR #1332, opened 2015-06-26
Make sure Essential information is not double tracked

drowka:
Don't run `trackView` on need-to-know views if they're already being tracked.


stevebrownlee:
:+1:

PR #1334, opened 2015-06-29
Fs update

finnursig:
variables and helpers added


finnursig:
@jcreamer898 approves!

PR #1339, opened 2015-07-07
Remove floor function from getMaxRangeValue method.

Keqi:
Floor function on number in range [0,1] returns always 0 or 1 which is not valid max value for trips.


tkali:
:+1:

PR #1340, opened 2015-07-07
Fixes the logo sponsor logo placement appearing on top of another add

benfalk:


gcziprusz:
Do it like this 
screen shot 2015-07-07 at 12 33 12 pm

gcziprusz:
then its :+1:

PR #1342, opened 2015-07-09
show notification to destinations-next alpha testers

gcziprusz:
Notification :point_down:  should show up if a cookie called `destinations_next_cookie` is set
screen shot 2015-07-08 at 8 19 35 pm


mkunkel:
LGTM but :x: above

PR #1343, opened 2015-07-14
Datepicker tweaks

WunderBart:


app/assets/javascripts/lib/components/datepicker.js
```diff
@@ -12,7 +12,6 @@ define([ "jquery", "picker", "pickerDate", "pickerLegacy" ], function($) {
     callbacks: {},
     dateFormat: "d mmm yyyy",
     dateFormatLabel: "yyyy/mm/dd",
-    listener: "#js-row--content",
```
robuye: what's up with this? has it never been used?

WunderBart: Yeah, I think someone copied it along with some component template (or from other component). Or maybe listener was used some time before, in previous versions, dunno - doesn't matter. Now it's not used anywhere.

PR #1346, opened 2015-07-14
Remove border bottom from last  of type

JoeShep:
Fixes issue where an extra bottom border line was being drawn when tours component was not present. This universally removes any bottom border from a component that is the last one before the footer.

Addresses this card: https://lonelyplanet.atlassian.net/browse/DCV-289


gcziprusz:
:8ball:  :shipit:

PR #1349, opened 2015-07-19
Remove loading indication from autocomplete

WunderBart:
This decreases CPU usage significantly (reported as a bug). Tried to replace it with something less demanding (-_-) but ended up deleting it. It wasn't of much use anyways - results are fetched before it's fully appeared. 

:fire: :fire: :fire:


nidalaa:
omg! such low CPU for FF :+1:

PR #1351, opened 2015-07-22
Am adblock tracker

amccadams:
This sends an event to Google Analytics when adblock in being used on a page.
To check:
Turn on Adblock in your browser.  
Open up console and type:
`window.lp.isAdblockActive`

If it is active should return true if it is inactive it should return false

 Note:  When I was working on this locally I could only get ads to load in the landing pages - hotels & car-rental
Jira Ticket
https://lonelyplanet.atlassian.net/browse/LPMC-105


JoeShep:
kewl. :+1

PR #1352, opened 2015-07-27
adds advertising.js to head manually so it will get loaded

amccadams:
This will make sure that the `advertising.js` file gets loaded in the head and will allow true to be set and fire off the GA event if ads are being blocked.


jcreamer898:
@WunderBart @bentedder @leejones, can you guys see any issues with this approach?

app/views/layouts/partials/snippets/_head_js.haml
```diff
@@ -85,6 +85,8 @@
 
 = yield :head_js
 
+%script{ src: "https://assets.staticlp.com/advertising.js"}
```
bentedder: do you think it needs to be above the `= yield :head_js` stuff in case any apps require their own rolled js files that include app_core?

WunderBart: Apart from what Ben said I don't think there should be any issues.

PR #1356, opened 2015-07-30
Tk fixing adds

tkali:
This is a small workaround that will prevent showing up ad placeholder before it's loaded. This needs to be part of more solid solution (I'm working on it).


robuye:
does it work?

tkali:
It won't catch the case where ads is loaded successfully, but still doesn't display it's content. I'm not sure how big of a problem is that. Need to see it on prod. 

Anyway this is just a dirty fix. We need still to deal with SSL conflicts, synced scripts, adblock...

robuye:
:+1:

PR #1361, opened 2015-08-07
Create special app_core for gonzo without ad_manager and advertising lib

drowka:
Ad_manager script from rizzo breaks the loading of js scripts in gonzo when the Adblock is turned on. There aren't any ads server in gonzo anyway, so we decided to create a special app_core for gonzo that doesn't require these two rizzo libraries.


amccadams:
:ship: :monkey:

PR #1362, opened 2015-08-10
Don't expand ads on load error and with empty placeholder

tkali:


nidalaa:
tested in community for a while- the result was pretty nice- placeholder wasn't displayed for failed ads :D
but it wasn't displayed in situation when there was no ads errors in console neither :trollface: 
keep digging @tkali

nidalaa:
Brand new solution is here https://github.com/lonelyplanet/rizzo/pull/1364/files, so I am closing it :D

PR #1363, opened 2015-08-17
Dont overwrite pageName when tracking flights widget click event

drowka:
`pageName` param in `flights_omniture` script shouldnt be overwritten when the` pageName` value is already present. This messed up statistics for flights widget in Essential Information
Jira ticket: https://lonelyplanet.atlassian.net/browse/LPMC-70


tkali:
:+1:

PR #1364, opened 2015-08-21
Fix expanding and collapsing od ad row in header

tkali:


app/assets/stylesheets/core/layouts/modern/_ads.sass
```diff
@@ -3,20 +3,23 @@
 // ----------------------------------------------------------------------------
 
 .row--sponsored
-  transition: height 0.5s ease
-  min-height: 60px
   overflow: hidden
+  @extend %transition--base
+
   &.row--leaderboard
     transition: height 0.5s ease, transform 0.4s ease
```
WunderBart: Do we need this transition at all? It should work without those 2 lines above, could you check it please? :]

app/assets/stylesheets/core/layouts/modern/_ads.sass
```diff
@@ -3,20 +3,23 @@
...
 
-  +respond-to(728)
-    min-height: 100px
-
   &.is-closed
```
WunderBart: You could get rid of `is-closed` class & move those styles below straight to the `row--sponsored` class as a default. Then you could just toggle `is-open` class in JS, if that makes sense.

app/assets/stylesheets/core/layouts/modern/_ads.sass
tkali: Yeah, I was thinking about removing it, but then I thought it may be used in community. Will rm.

app/assets/stylesheets/core/layouts/modern/_ads.sass
tkali: Sound like a good idea.

WunderBart: \o/

tkali: _I'm afraid I cannod do that Dave_

tkali: OK, from what I see only community changes `min-height`, we can remove it from there and use your way of expanding row.

nidalaa:
@WunderBart  @tkali taking into account a discussion here: https://github.com/lonelyplanet/community/pull/2389. What about other apps, which are using Rizzo(e.g. Shop)? Do they need some custom changes as well?
_Remember that in Shop Rizzom bump is automated_

tkali:
@nidalaa pls... I'm a blackthumb I know my engine.



_it should be fine_

nidalaa:
_it should be fine_ along with _what can possibly go wrong_ do not sounds safe for me ;)
![Uploading What+could+possibly+go+wrong_dd55c4_3966855.gif…]

tkali:
I greped through my repos and if they touch `row--sponsored` it's only to set it to `display: none`.

No risk no fun. You can always make rizzo solution community-centric but I guess that's not what rizzo is for.

nidalaa:
Community production is working on this branch. It seems to be :+1: 
dance_of_joy

tkali:
Good.


Merging.

PR #1366, opened 2015-08-24
Bart add custom title to secondary nav

WunderBart:
It adds option to add custom title (html_safe) to be displayed instead of selected item. We'd like use it in Community to save some precious space. 

It's backwards compatible, won't blow anything anywhere up (I hope so :trolleybus:).

Before: 

screenshot 2015-08-25 00 53 30

After:

screenshot 2015-08-25 00 50 38


jcreamer898:
Seems legit.

PR #1369, opened 2015-09-04
adds html safe so form labels can include html

amccadams:
This adds .html_safe class to form labels.  Please check that html can be added in yml inputs and renders as such.


drowka:
:+1:

PR #1373, opened 2015-09-21
Set banner logic to false until we're ready

jcreamer898:
This will allow us to deploy this code without launching so we can do some testing.


edubkendo:
:+1:

leejones:
:+1:

sonasi:
:+1:

PR #1375, opened 2015-10-05
configure_js should merge objects instead overwriting.

robuye:
real source of `extend` function used (uglified in the source):

```
function extend (target, source) {
  target = target || {};
  for (var prop in source) {
    if (typeof source[prop] === 'object') {
      target[prop] = extend(target[prop], source[prop]);
    } else {
      target[prop] = source[prop];
    }
  }
  return target;
}
```


robuye:
/cc @WunderBart @tkali 

this makes our `utag_data` for tealium work as expected.

tkali:
:shipit: when :green_heart:

PR #1376, opened 2015-10-07
Apply Fixes to BetaHelper

benfalk:
Rizzo was not correctly working on `Hotels` after it was bumped to
the latest version of Rizzo.  The error was an unitialized constant
of PlacePresenter because Hotels lacks such a presenter.  This
insulates that condition and wraps the module with tests to prevent
further problems.


jcreamer898:
You sir, are a gentleman and a scholar.

PR #1378, opened 2015-10-07
Fix merge js

robuye:
full source of `extend`:

```
function extend (target, source) {
  target = target || {};
  for (var prop in source) {
    if((typeof source[prop] === 'object') && source[prop] !== null && source[prop].constructor == Array){
      target[prop] = source[prop];
    } else if (typeof source[prop] === 'object') {
      target[prop] = extend(target[prop], source[prop]);
    } else {
      target[prop] = source[prop];
    }
  }
  return target;
}
```

handles arrays and null values properly.

Also define namespace key if it's empty before merging. So when merge happens the key is always present.

### DIFF
--- a/app/assets/javascripts/lib/core/user_feed/content.js
+++ b/app/assets/javascripts/lib/core/user_feed/content.js
@@ -74,7 +74,8 @@ define([
 
     return this[type].$container
       .find(this.config.item)
-      .slice(0, latestCount);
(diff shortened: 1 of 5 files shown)


nidalaa:
This branch has been used in community for 3 days and everything works like it should :dancer: 
:shipit: @WunderBart

WunderBart:
:+1:

PR #1379, opened 2015-10-09
Prevent self notifications from popping up

WunderBart:
@nidalaa this should be just enough :dancer:


nidalaa:
Cherry-picked here https://github.com/lonelyplanet/rizzo/pull/1378/commits

PR #1381, opened 2015-10-14
Bump jquery-timeago to v1.4.3

WunderBart:
Should fix annoying memory leak causing mad CPU heat caused by stacking intervals on nodes that are no longer in DOM.
ref: https://github.com/rmm5t/jquery-timeago/commit/b5a364d3a24b4e69aaec2005e32649a494eed826


drowka:
Looks good to me!

WunderBart:
:sheep:-ing!

PR #1384, opened 2015-10-22
Fix JOIN button position in main nav

WunderBart:
From screenshot 2015-10-22 12 30 30 to screenshot 2015-10-22 12 30 09


robuye:
:+1:

PR #1386, opened 2015-10-23
Fix wrong target on feed item click

WunderBart:
This fixes issue with catching wrong elements and redirecting to `undefault` bug.


nidalaa:
much better in community now :+1:

PR #1388, opened 2015-11-06
Fix the wrong value of 'narrow-view' variable in documentation

drowka:
Proper `narrow-view` breakpoint value is `380px`


WunderBart:
:heart: :ship:

PR #1392, opened 2015-11-16
Fix for squashed MPU card on mobile

drowka:
This fixes issue when MPU card got squashed on devices with viewport width lower than 380px. 
JIRA ticket: https://lonelyplanet.atlassian.net/browse/LPMC-269

**Before**
Before image

**After**
After image

This doesn't affect the cards behaviour on larger screens.

Nexus 7 screenshot


WunderBart:
:clap: :ship:

PR #1394, opened 2015-11-19
Treat submenus of global navigation items as click-trackable.

tkali:
This one ads `data-lpa-` tracking attributes to submenus of core navigation items. Previously only primary nav items were taken into account. 

Pix for reference:

### DIFF
--- a/app/views/components/navigation/_submenu.html.haml
+++ b/app/views/components/navigation/_submenu.html.haml
@@ -2,7 +2,7 @@
   .nav--stacked.nav__submenu__content.js-submenu.icon--tapered-arrow-up--after.icon--white--after
     - if item
       - item[:submenu].each do |sub|
-        = render "components/navigation/nav_item", item: sub, active: false, submenu_item: true
+        = render "components/navigation/nav_item", item: sub, active: false, submenu_item: true, global: sub[:global]
 
     - if defined?(button_text) && button_text
       .nav__submenu__button
--- a/app/views/layouts/partials/inline_js/_nativo.html
+++ b/app/views/layouts/partials/inline_js/_nativo.html
@@ -1,1 +1,1 @@
-<script type="text/javascript" src="//a.postrelease.com/serve/load.js?async=true"></script>
+<script async type="text/javascript" src="//a.postrelease.com/serve/load.js?async=true"></script>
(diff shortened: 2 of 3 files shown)


app/views/layouts/partials/inline_js/_nativo.html
```diff
@@ -1 +1 @@
-<script type="text/javascript" src="//a.postrelease.com/serve/load.js?async=true"></script>
+<script async type="text/javascript" src="//a.postrelease.com/serve/load.js?async=true"></script>
```
WunderBart: Does it need to have both `async` attribute & `async=true` param?

app/views/components/navigation/_submenu.html.haml
```diff
@@ -2,7 +2,10 @@
   .nav--stacked.nav__submenu__content.js-submenu.icon--tapered-arrow-up--after.icon--white--after
     - if item
       - item[:submenu].each do |sub|
-        = render "components/navigation/nav_item", item: sub, active: false, submenu_item: true
+        - if sub[:global]
+          = render "components/navigation/nav_item", item: sub, active: false, submenu_item: true, global: true
+        - else
+          = render "components/navigation/nav_item", item: sub, active: false, submenu_item: true
```
WunderBart: Seems like 
``` haml
= render "components/navigation/nav_item", item: sub, active: false, submenu_item: true, global: sub[:global]
```
would be enough?

app/views/layouts/partials/inline_js/_nativo.html
tkali: I don't know about `async=true` it may mean anything on the vendor's side. It may inject script that builds `<script>` tag for async payload download or smth. Anyway `async` _attribute_ here helps regardless.

app/views/components/navigation/_submenu.html.haml
tkali: Unfortunately not, because I check for `defined?(global)` (it's tricky man) inside `nav_item` partial, so you could pass `global: false` here and it still would yield `defined?(global) == true`.

WunderBart: What about `defined(global) && global`?

app/views/layouts/partials/inline_js/_nativo.html
WunderBart: ok cool :+1:

WunderBart:
Noice. Now squasit & :shipit: :3

PR #1395, opened 2015-11-20
Change Destinations Next beta opt-in/out banner styling

JoeShep:
Updates alert styling to use custom background and text alignment specific to the beta test opt-in/out alert


jcreamer898:
Nice.

PR #1397, opened 2015-11-22
Move prevent default

jcreamer898:
The prevent default needs to be higher up to make sure it gets cancelled before the location reload.


WunderBart:
:+1:

PR #1399, opened 2015-11-24
Beta banner load w/ JavaScript

jcreamer898:
Going to show the beta banner w/ JavaScript now instead of Ruby because of caching issues.


benfalk:
Looks good, maybe squash these two commits into a single commit and provide a better git commit message :+1: 

http://benfalk.com/blog/2015/10/20/commits-that-matter/ 
 * The Old One Two
 * See Diff for Similar Message

PR #1402, opened 2015-11-24
Remove ads in tests

jcreamer898:
Since we pass `ad_config`'s from all over the site to this single method, this might be the one place we can test for ads and decide to not show them in the context of acceptance tests. I sure hope so at least...


jcreamer898:
:+1: from @mkunkel

PR #1405, opened 2015-11-25
Don't load ads in test environment

tkali:
This one prevents from initialising `AdManager` in testing environments.


jcreamer898:
Good deal.

tkali:
Alright then, let's summon @WunderBart

WunderBart:
looks good :+1:

PR #1406, opened 2015-11-25
Remove async attribute

jcreamer898:
As of 6 days ago when we put `async` on this, something odd started happening and a bunch of errors started randomly appearing...

I think what's happening is the Nativo script is also using a `umd` module for jQuery. When it loads async, it sometimes overwrites our own jquery and they happen to use a version of jQuery that does not support the `.wrap` method for whatever reason. Not to mention it's `2.1.1`.

Let's remove the async and see if the errors drop off.


tkali:
That's my bad.

PR #1407, opened 2015-11-26
Manage url params in controller.js

nidalaa:
PROBLEM: So far empty params have been ALWAYS overwritten by current window.location.search. Probably that was a reason of introducing removePageParam here: https://github.com/lonelyplanet/rizzo/commit/cf3e10ceb2464819d4848272186b2e20de0a7483. Because of this, after closing lightbox page param diseappeared(problem spotted in community) while an other params were added to lightbox url. 

SOLUTION: Every time some state is pushed to local variable in _generateState both: url and params are provided and empty params are not overwritten by params from window.location(so, we respect that in some cases params do not exist).

Tested locally on community and gonzo.

/cc @WunderBart @drowka

### DIFF
--- a/app/assets/javascripts/lib/page/controller.js
+++ b/app/assets/javascripts/lib/page/controller.js
@@ -28,7 +28,7 @@ define([
 
   Controller.prototype.init  = function() {
     this.pushState = new PushState;
-    this._generateState(this.getSlug());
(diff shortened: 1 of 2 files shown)


WunderBart:
:shipit:

PR #1408, opened 2015-11-26
Use working 'antarctica' slug

tkali:
This one brings working slug for "Antarctica" place. 

Related: https://github.com/lonelyplanet/destinations-next/pull/434


WunderBart:
What's that number? How does it bring Antarctica back? :O
(Just asking for some explanation)

nidalaa:
Long story short @WunderBart : it should be redirected in dispatch but ops are not here. Therefore temp solution in rizzo

https://github.com/lonelyplanet/destinations-next/pull/434

PR #1409, opened 2015-11-27
Make sure presenters are defined before using them in BetaHelper

tkali:
This fixes case when an app that uses rizzo has no `PlacePresenter` and no `ArticlesShowPresenter` defined.


WunderBart:
:+1:

PR #1411, opened 2015-12-03
Tests for changes in history.state without using spies

nidalaa:
Change was forced after jasmine-core update to 2.4.0, more information: https://github.com/jasmine/jasmine/issues/948

----------------------
@WunderBart  @jcreamer898 can you take a look?

_I was also thinking about adding `expect(history.state).toEqual({});` but history.state is supported in phantomjs 2.0 and well it's additional thing to test. The most important is covered \o/_

### DIFF
--- a/spec/javascripts/lib/components/gallery_spec.js
+++ b/spec/javascripts/lib/components/gallery_spec.js
@@ -50,15 +50,21 @@ define([ "jquery", "public/assets/javascripts/lib/components/gallery.js" ], func
     });
 
     describe("Updating the history", function() {
-      var newSlug = "my-image";
+      var initialSlug = null,
+          newSlug = "my-image";
+
       beforeEach(function(){
-        spyOn(window.history, "pushState");
+        initialSlug = window.location.pathname;
         gallery._updateSlug(newSlug);
       });
 
       it("Updates the state", function() {
         var expectedSlug = $("#js-gallery").data("href") + "/" + newSlug;
-        expect(window.history.pushState).toHaveBeenCalledWith({  }, "", expectedSlug);
+        expect(window.location.pathname).toEqual(expectedSlug);
+      });
+
+      afterEach(function() {
+        window.history.replaceState({}, null, initialSlug)
       });
     });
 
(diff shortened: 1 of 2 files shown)


spec/javascripts/lib/components/gallery_spec.js
```diff
@@ -50,15 +50,21 @@ define([ "jquery", "public/assets/javascripts/lib/components/gallery.js" ], func
...
       });
 
       it("Updates the state", function() {
         var expectedSlug = $("#js-gallery").data("href") + "/" + newSlug;
-        expect(window.history.pushState).toHaveBeenCalledWith({  }, "", expectedSlug);
+        expect(window.location.pathname).toEqual(expectedSlug);
+      });
+
+      afterEach(function() {
+        window.history.replaceState({}, null, initialSlug)
```
WunderBart: I think it would be safer just to backup whole `window.history` in `before` and restore it in `after`. Like dis:

```javascript
beforeEach(function() {
  var tempHistory = window.history;
  ...
}

afterEach(function() {
  window.history = tempHistory;
}
```

What do you think, @nidalaa?

spec/javascripts/lib/page/pushstate_spec.js
```diff
@@ -95,13 +95,20 @@ define([ "jquery", "public/assets/javascripts/lib/page/pushstate.js" ], function
...
       });
-      it("history.pushState is called", function() {
-        expect(history.pushState).toHaveBeenCalledWith({}, null, "/test");
+
+      it("changes current location", function() {
+        expect(window.location.pathname).toEqual("/test");
+      });
+
+      afterEach(function() {
+        window.history.replaceState({}, null, initialSlug)
```
WunderBart: Same as above.

spec/javascripts/lib/components/gallery_spec.js
jcreamer898: Will that *really* do anything? `window.history` is an object. so `tempHistory` is just a reference to it. So it **should** be exactly the same...

WunderBart: Oh, yeah... :facepalm: Just ignore me - have to take a break :stuck_out_tongue: 

...unless you want to use deep copy :trollface: 

```javascript
tempHistory = $.extend(true, {}, window.history);
```

WunderBart:
:shipit: :trollface:

PR #1412, opened 2015-12-10
Call trackEvent directly instead of tracking page view in Controller

tkali:
This one makes Controller send "page view" event to analytics tracker for each virtual page update  instead of relying on `dataLayer` and `lp.analytics.api.trackPageView`.


WunderBart:
:+1:

PR #1415, opened 2015-12-16
Mk newsletter form component

laserchicken:
Newsletter endpoint changed (reimplemented from scratch in gonzo) and footer form in rizzo stopped working. New endpoint responds with http status codes and this component is responsible for handling these responses after submitting the form.
https://lonelyplanet.atlassian.net/browse/LPMC-302

### DIFF
--- a/app/assets/javascripts/lib/core/newsletter_form.js
+++ b/app/assets/javascripts/lib/core/newsletter_form.js
@@ -45,7 +45,7 @@ define([
       content: "Thank you for subscribing, " +
         "you'll soon receive an email confirming your subscription."
     }, true);
-  }
+  };
 
   NewsletterForm.prototype._handleSubmitError = function(xhr) {
     if (xhr.status === 409) {
@@ -59,7 +59,7 @@ define([
         content: "Something went wrong."
       }, true);
     }
-  }
+  };
 
   $(document).ready(function() {
     new NewsletterForm;
--- a/app/assets/javascripts/lib/core/newsletter_form.js
+++ b/app/assets/javascripts/lib/core/newsletter_form.js
@@ -14,7 +14,7 @@ define([
   // el: {string} selector for form
   // alert: {string} selector for alert wrapper,
   //        will display notifications after submit,
-  //        by default in place of submitted form
+  //        by default in place of submitted form (see defaults)
   function NewsletterForm(args) {
     this.config = $.extend({}, defaults, args);
     this.$el = $(this.config.el);
@@ -23,7 +23,7 @@ define([
   }
 
   NewsletterForm.prototype.init = function() {
-    this.alert = new Alert({ container: defaults.alert });
+    this.alert = new Alert({ container: this.$alert });
     this.listen();
   };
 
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/lib/core/newsletter_form.js
```diff
@@ -0,0 +1,67 @@
...
+  //        by default in place of submitted form
+  function NewsletterForm(args) {
+    this.config = $.extend({}, defaults, args);
+    this.$el = $(this.config.el);
+    this.$alert = $(this.config.alert);
+    this.$el.length && this.init();
+  }
+
+  NewsletterForm.prototype.init = function() {
+    this.alert = new Alert({ container: defaults.alert });
```
WunderBart: Use `this.config` instead of `defaults` var, just like in line 21 :smile:

app/assets/javascripts/lib/core/newsletter_form.js
```diff
@@ -0,0 +1,67 @@
...
+
+    var that = this;
+
+    $.post(this.$el.attr("action"), this.$el.serialize())
+      .done(function() {
+        that.alert.success({
+          title: "Success!",
+          content: "Thank you for subscribing, " +
+            "you'll soon receive an email confirming your subscription."
+        }, true);
```
WunderBart: Could you move the above to separate function please? `_handleSubmitSuccess` would be perfect.

app/assets/javascripts/lib/core/newsletter_form.js
```diff
@@ -0,0 +1,67 @@
...
+            title: "",
+            content: "You are already subscribed."
+          }, true);
+        } else {
+          that.alert.error({
+            title: "Error. ",
+            content: "Something went wrong."
+          }, true);
+        }
+      });
```
WunderBart: Same as above - `_handleSubmitError`

app/assets/javascripts/lib/core/newsletter_form.js
```diff
@@ -0,0 +1,67 @@
+define([
+  "jquery",
+  "lib/utils/alert"
+], function($, Alert) {
+
+  "use strict";
+
+  var defaults = {
+    el: ".js-newsletter-footer",
+    alert: ".js-newsletter-footer"
```
WunderBart: I think there's no need to make another setting for alert if the target is the same. Just use `$el` everywhere and make it super simple :+1:

app/assets/javascripts/lib/core/newsletter_form.js
WunderBart: If you use `$el` everywhere, you can actually pass `this.$el` here instead of `this.config.el`.

WunderBart:
Just few minor comments. Overall it looks really good :+1:

app/assets/javascripts/lib/core/newsletter_form.js
laserchicken: the idea was to make this configurable: if developer doesn't want default behaviour, which is display alert in place of form, he can pass alert wrapper and display it there, without hiding the form

WunderBart: Cool :+1:

WunderBart:
Never merge without +1 at the very end of your work, please.

PR #1417, opened 2015-12-21
Change flights widget's tracking event params

tkali:
This is a tweak of params we send for tracking when user submits search form in Flights widget.


tkali:
looks good to me :trollface: 
:+1:

PR #1418, opened 2015-12-21
Mk newsletter form component

laserchicken:


app/models/sailthru_email.rb
```diff
@@ -0,0 +1,13 @@
+class SailthruEmail
+  SEND_EMAIL_URL = "http://canary.community.lonelyplanet.com/sailthru/v1/email"
```
robuye: please use ENV instead of hardcoding URL. We will change it at some point. The ENV needs to be defined in the parent application so it's best to ensure it's present somehow.

app/helpers/sailthru_helper.rb
```diff
@@ -26,4 +26,16 @@ def sailthru_tags(opts = {})
       haml_tag(:meta, name: "sailthru.tags",  content: tags) if tags
     end
   end
+
+  def register(params)
```
robuye: I don't like it here. Helpers should be just view helpers, I'd rather see it as proper class.

app/models/sailthru_user.rb
```diff
@@ -0,0 +1,35 @@
+class SailthruUser
+  REGISTER_OR_CHECK_URL = "http://canary.community.lonelyplanet.com/sailthru/v1/users"
```
robuye: this can't be hardcoded here, it has to be an ENV.

app/models/sailthru_user.rb
```diff
@@ -0,0 +1,35 @@
...
+  SAILTHRU_REGIONS = YAML.load(File.read(File.expand_path('../../data/sailthru/regions.yml', __FILE__)))
```
robuye: I'd rather build a path using `Rizzo::Engine.root` instead of using relative paths `../..`.

app/models/sailthru_email.rb
```diff
@@ -0,0 +1,13 @@
+class SailthruEmail
```
robuye: I'd wrap Email and User into one class, feels small enough to handle all responsibilities of talking to Sailthru without sacrificing readability. Plus I don't expect the class to grow in the future as it pretty much covers everything we intend to do with Sailthru.

app/controllers/rizzo/sailthru_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Rizzo::SailthruController < ActionController::Base
+  include SailthruHelper
```
robuye: this is bad, you're including view helpers to controller. Let's use Helpers strictly for views. Extract the method in helper to separate class. Please namespace it with `Rizzo`.

WunderBart:
Looks good to me :+1:
Squash if possible & :shipit:, unless @robuye says otherwise :trolleybus:

app/controllers/rizzo/sailthru_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Rizzo::SailthruController < ActionController::Base
+
+  def save
+    render json: {}, status: SailthruService.register(form_params)
+  end
+
+  private
+
+  def form_params
+    params.require(:sailthru)
```
robuye: does it work without `permit`?

robuye:
can you add tests please?

app/controllers/rizzo/sailthru_controller.rb
laserchicken: sure, I'm not doing mass assignment anywhere

laserchicken:
@robuye is this ok? can I merge it?

robuye:
:+1: sorry for the delay.

PR #1423, opened 2016-01-11
Switch label <-> action in event params for booking widgets

tkali:
So, there was an error in the spec that slipped into codebase. `action` you see here fellas should be `label`. Yeeup.


WunderBart:
:shipit:

PR #1424, opened 2016-01-11
Simplify tracking params of car-rental widget

tkali:
This one makes car rental forms that appear aside on some NTK pages pass Tealium requirements.


WunderBart:
:+1:

PR #1426, opened 2016-01-11
Including rizzo next npm dist folder

jcreamer898:
When gems get copied down they exclude what's in the gitignore so rizzo next was getting left out.


jcreamer898:
:+1: from @laserchicken

PR #1428, opened 2016-01-12
Adding only the dist folder to the paths

jcreamer898:
I think the problem was there were a bunch of files making into the asset path from other `node_modules`. Adding only what's needed seems to help.


laserchicken:
:+1:

PR #1429, opened 2016-01-12
Add rizzo next to precompile, update to named AMD module

jcreamer898:
Rizzo Next needed to be updated to be included as a named module so that it wouldn't break when it actually ran in the browser. Also needed to add rizzo-next.css to the precompile so it gets deployed to s3.


jcreamer898:
:+1: from @laserchicken

PR #1430, opened 2016-01-12
Manage url params in controller.js

nidalaa:
PROBLEM: So far empty params have been ALWAYS overwritten by current window.location.search. Probably that was a reason of introducing removePageParam here: https://github.com/lonelyplanet/rizzo/commit/cf3e10ceb2464819d4848272186b2e20de0a7483. Because of this, after closing lightbox page param diseappeared(problem spotted in community) while an other params were added to lightbox url. SOLUTION: Every time some state is pushed to local variable in _generateState both: url and params are provided and empty params are not overwritten by params from window.location(so, we respect that in some cases params do not exist)

### DIFF
--- a/spec/javascripts/lib/page/controller_spec.js
+++ b/spec/javascripts/lib/page/controller_spec.js
@@ -86,7 +86,7 @@ define([
         expect(controller.states[controller.states.length - 1].state).toEqual(deserialized);
       });
 
-      it("updates the application state object empty object if paramseters are not provided", function() {
+      it("updates the application state object empty object if parameters are not provided", function() {
         controller._generateState("url", null);
         expect(controller.states[controller.states.length - 1].state).toEqual({});
       });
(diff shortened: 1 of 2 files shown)


nidalaa:
Last time, this P/R had to be reverted after fails in Waldorf. After debugging and last changes, tests for Waldorf are passing (https://github.com/lonelyplanet/waldorf/pull/1975) \o/.

@WunderBart I would be happy to merge it after global nav will be stable, what do you think?

spec/javascripts/lib/page/controller_spec.js
```diff
@@ -81,10 +81,15 @@ define([
...
       });
 
-      it("updates the application state object with the search parameters", function() {
-        controller._generateState();
+      it("updates the application state object with provided parameters", function() {
+        controller._generateState("url", controller.getParams());
         expect(controller.states[controller.states.length - 1].state).toEqual(deserialized);
       });
+
+      it("updates the application state object empty object if paramseters are not provided", function() {
```
WunderBart: HA, type-o! :trollface:

WunderBart:
Looks good! :+1:

nidalaa:
After applying global nav changes and additional tests in communit, waldorf and hotels- merging finally!

PR #1432, opened 2016-01-12
Remove files causing precompile to fail

jcreamer898:
Since there's ES6 in the build one, it was bombing out during some apps' precompile step.


laserchicken:
:+1:

PR #1442, opened 2016-01-15
Make UserFeed an independent component

WunderBart:
- remove from core components (make optional)
- leave unread messages count (old mobile nav) in authenticator.js
- remove feed from user menu partial completely
- make separate styles for what’s left (_user_menu.sass)
- clean up all the things


nidalaa:
:+1: branch tested in both Community and Luna, with nev global nav and without

:shipit:

PR #1443, opened 2016-01-15
Bump Rizzo Next

jcreamer898:
Pulling in change from most recent RN.
Had to rename the cookie check for global nav. Dispatch adds an underscore to it.


thomasthesecond:
:+1:

PR #1444, opened 2016-01-16
namespace fix; form action path fix

laserchicken:
This change fixes `NameError (uninitialized constant Rizzo::SailthruController::SailthruService)`. Calling SailthruSevice from Rizzo::SailthruController was throwing NameError due to improper nesting (see. http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#nesting). I decided to use module nesting convention instead and it fixed the problem.

### DIFF
--- a/app/controllers/rizzo/sailthru_controller.rb
+++ b/app/controllers/rizzo/sailthru_controller.rb
@@ -1,7 +1,7 @@
 class Rizzo::SailthruController < ActionController::Base
 
   def save
-    render json: {}, status: SailthruService.register(form_params)
+    render json: {}, status: Rizzo::SailthruService.register(form_params)
   end
 
   private
--- a/app/controllers/rizzo/sailthru_controller.rb
+++ b/app/controllers/rizzo/sailthru_controller.rb
@@ -1,12 +1,14 @@
-class Rizzo::SailthruController < ActionController::Base
+module Rizzo
+  class SailthruController < ActionController::Base
 
-  def save
-    render json: {}, status: Rizzo::SailthruService.register(form_params)
-  end
+    def save
+      render json: {}, status: SailthruService.register(form_params)
+    end
 
-  private
+    private
 
-  def form_params
-    params.require(:sailthru)
+    def form_params
+      params.require(:sailthru)
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


app/views/layouts/partials/snippets/_footer_sitemap.html.haml
```diff
@@ -12,7 +12,7 @@
         %strong
           20% discount
         on your next guidebook purchase
-      %form.newsletter-footer.js-newsletter-footer(action='//www.lonelyplanet.com/rizzo_sailthru' method='post')
+      %form.newsletter-footer.js-newsletter-footer(action='/rizzo_sailthru' method='post')
```
robuye: what's the difference between this and previous version?

app/controllers/rizzo/sailthru_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Rizzo::SailthruController < ActionController::Base
 
   def save
-    render json: {}, status: SailthruService.register(form_params)
+    render json: {}, status: Rizzo::SailthruService.register(form_params)
```
robuye: Why is this necessary? Won't rails lookup `SailthruService` within `Rizzo` namespace anyways?

app/views/layouts/partials/snippets/_footer_sitemap.html.haml
laserchicken: no difference really, just a shorter version

app/controllers/rizzo/sailthru_controller.rb
laserchicken: sometimes I get `NameError (uninitialized constant Rizzo::SailthruController::SailthruService)`, after adding this rails lookup seems to work properly

laserchicken:
failing build is not related to this change and is fixed https://github.com/lonelyplanet/rizzo/commit/e99ebb130850e39f3f3f147ee7d2040842939b32

tkali:
:+1:

app/views/layouts/partials/snippets/_footer_sitemap.html.haml
nidalaa: @laserchicken  what about apps with different domain? `shop.lonelyplanet.com` or `auth.lonelyplanet.com`? Would it work?

nidalaa: ooops, too late... anyway I summon you @laserchicken

PR #1445, opened 2016-01-18
Make global header match Rizzo Next

thomasthesecond:
Because Rizzo Next uses rems and a base font size 10px and Rizzo uses a base font size of 14px, overrides needed to be added to make the header look the same across all areas of the site.

[View Codepen demo](http://codepen.io/thomasthesecond/pen/977f587e591f7d06d78f40f149bd6ad1)

### DIFF
--- a/app/assets/stylesheets/core/layouts/modern/_primary_nav_new.sass
+++ b/app/assets/stylesheets/core/layouts/modern/_primary_nav_new.sass
@@ -68,8 +68,9 @@ $lp-min-480: #{(480 * .0625)}em   // 30em
 $lp-min-360: #{(360 * .0625)}em   // 22.5em
 $lp-min-320: #{(320 * .0625)}em   // 20em
 
-$lp-max-720: #{(719 * .0625)}em // 44.9375em
-$lp-max-480: #{(479 * .0625)}em // 29.9375em
+$lp-max-1530: #{(1530 * .0625)}em // 95.625
+$lp-max-720: #{(719 * .0625)}em   // 44.9375em
+$lp-max-480: #{(479 * .0625)}em   // 29.9375em
 
 $lp-blue: #287bbb
 $lp-color-red: #dc221a
@@ -277,11 +278,17 @@ $lp-notification-badge-size: 17px
     top: (10rem * $scaling-factor)
     font-size: (1.8rem * $scaling-factor)
 
+    @media (max-width: $lp-max-1530)
+      right: 0
+
     &:before
       border: (.8rem * $scaling-factor) solid transparent
       border-bottom-color: #fff
       margin-left: (-.8rem * $scaling-factor)
       top: (-1.6rem * $scaling-factor)
+
+      @media (max-width: $lp-max-1530)
+        right: (1.6rem * $scaling-factor)
 
   .sub-navigation--visible
     box-shadow: 0 $lp-gutter (9rem * $scaling-factor) rgba(0, 0, 0, .4)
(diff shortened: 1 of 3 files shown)


thomasthesecond:
Here's my note about using `!important`:

> Search and mobile navigation need !important because rizzo-next.css loads inside of the `body` and therefore after Rizzo's styles. The `content-wrapper navigation-wrapper rizzo-navigation rizzo-next` class chain cannot be used for these components because they're appended to the `body` and not the `header`.

I don't think we can load rizzo-next.css in the `head` and before Rizzo's CSS, but maybe there's a way to add a class to the search and mobile-navigation components *only* when they load alongside Rizzo? Another option would be to append the components to `header` instead of `body` but there would probably some CSS rework. This would be a question for @jcreamer898.

nidalaa:
Regarding failing spec on jenkins: changes in user feed have been merged recently. Merging master changes could be the easiest way to handle failing specs in this branch :D

thomasthesecond:
Fixed, thanks!

nidalaa:
I think we should get rid of that red dot entirely.

Reasons:
- it was introduced to design when we were using "old" user feed, without that feed this indicator is not self- explanatory
- currently we have side feed(enabled only in community)
image

image

which indicates when message inside user feed wasn't viewed. In the same time, red dot in global nav stays there until user opens the message itself. We would produce inconsistent information for forum users(and basically they are users, who can have messages).

nidalaa:
:point_up:  //cc @bpotterbaum @jcreamer898

jcreamer898:
@thomasthesecond we could change where the `rizzo-next.css` loads, it currently just comes from... https://github.com/lonelyplanet/rizzo/blob/master/app/views/layouts/custom/_post_header.html.haml#L6

@nidalaa don't have to convince me, I'm all about nuking it

thomasthesecond:
Thanks @jcreamer898. I'll look into that.

thomasthesecond:
Related: https://github.com/lonelyplanet/rizzo-next/pull/320

WunderBart:
I've noticed quiet a few places that could use the [`size`](https://github.com/lonelyplanet/rizzo/blob/master/app/assets/stylesheets/sass/mixins/_utilities.sass#L29)& [`font-size`](https://github.com/lonelyplanet/rizzo/blob/master/app/assets/stylesheets/sass/mixins/_utilities.sass#L6) mixins. I'd suggest using [`z-layer`](https://github.com/lonelyplanet/rizzo/blob/master/app/assets/stylesheets/sass/mixins/_utilities.sass#L44) one too, if possible. 

Just my 2 cents here :smile:

nidalaa:
I've tried to test it locally in community and luna but I cannot see changes from https://github.com/lonelyplanet/rizzo-next/issues/319.

I switched to local Rizzo, bumped rizzo-next version in package.json, installed new version and nothing :(
Am I missing something in my setup @thomasthesecond @WunderBart ?

WunderBart:
Npm package version needs to be bumped first. [Last one](https://github.com/lonelyplanet/rizzo-next/commit/5c1d0a748ad677238b9e168b969bcd907f418672) didn't include latest changes.

jcreamer898:
I'll do that as soon as I get in this am. 

Sent from Outlook Mobile




On Thu, Jan 21, 2016 at 6:09 AM -0800, "Bart Kalisz" <notifications@github.com> wrote:












Npm package version needs to be bumped first. Last one didn't include latest changes.

thomasthesecond:
@WunderBart 

`size` mixin has been used where I have width and height

`font-size` mixin won't work in this case because of the `$scaling-factor`. Rizzo uses a base of 14 and Rizzo Next a base of 10; I need to multiply the rem font size from Rizzo Next by that ratio in order to get the correct font size. The rem value created by the `font-size` mixin is too large, so I just created a custom mixin called `lp-font-size` that's similar but specific to this case.

`z-layer` mixin won't work for me because I have `z-index: 2` in two places, but the mixin doesn't accept custom values. I could extend it, but not sure it's worth it in this case.

WunderBart:
It's perfect :+1:

PR #1447, opened 2016-01-20
Make UserFeed self-instantiate

WunderBart:


nidalaa:
perfecto, muito obragada :+1:

PR #1448, opened 2016-01-21
prefix rizzo_sailthru path with app name

laserchicken:


robuye:
I don't think adding dispatch dependency to the newsletter form is good choice, it's too difficult to maintain over time :-1:

laserchicken:
ok, closing it

PR #1453, opened 2016-01-24
specify full url for newsletter submit endpoint

laserchicken:
This will also work for subdomains like shop.lonelyplanet.com.


robuye:
:+1:

PR #1454, opened 2016-01-25
Use pixel values for global header overrides

thomasthesecond:
This fixes a sizing issue with the new Rizzo Next global header in the shop. The issue is that the sizing is too big due to the use of `rem` values in Rizzo Next.

Rizzo Next uses a base font size of `10px` for rems, Rizzo uses `.875em` and the shop uses `100%`. Removing the scaling factor ratio and using `px` for values fixes the sizing problem on the shop. We don't want to alter the base font size of the shop because that could have unintended side effects.

In case this issue were to happen somewhere else, using `px` provides consistent rendering across all apps.


thomasthesecond:
Another (related) issue is that the icon fonts from Rizzo Next aren't loading in the shop. See 47f550f above for more details.

thomasthesecond:
Rizzo Next JavaScript isn't loading in the shop. See a674a1c above for details.

WunderBart:
Looks legit :+1:

PR #1455, opened 2016-01-26
load sailthru form in app_core_legacy

laserchicken:
Load sailthru form handling js in app_core_legacy (for apps like shop).


thomasthesecond:
:+1:

PR #1457, opened 2016-01-28
[DO NOT MERGE]Throwback thursday

WunderBart:
After latest changes, we cannot get UserFeed to work in production env. Locally - everything is fine. Couldn't find source of the issue so we're going back to before-rizzo-next state & cherry-pick changes we need to have in Community. Let's see if that works :trollface:


nidalaa:
Not needed anymore.

PR #1460, opened 2016-01-28
Add interest to dfp targeting

jcreamer898:
Not sure why, but interest was never sent to DFP for targeting...


thomasthesecond:
:+1:

PR #1462, opened 2016-02-02
Setup airbrake only for v4

robuye:
There is a breaking change in Airbrake API v5. If an application tries to use
v5 it will crash in rizzo initializer. This means Rizzo locks out every application
from being ever updated (just the airbrake).

Skip airbrake config if an application uses v5.

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -1,5 +1,15 @@
 if Rails.env.production? && defined?(Airbrake)
-  Airbrake.configure do |config|
-    config.api_key = ENV['AIRBRAKE_API_KEY']
+  airbrake_version = Gem::Version.new(Airbrake::VERSION)
+  new_version = Gem::Version.new('5.0')
+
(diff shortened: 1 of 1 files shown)


nidalaa:
> /mnt/workspace/Rizzo-Branch/config/initializers/airbrake.rb:10: Invalid return (SyntaxError)

----------------
So, in case of Airbrake 5 we are going to depend on specific app config, right?
Maybe we can change it and skip airbrake config in Rizzo if it is defined in app? Would it be possible?

robuye:
Rizzo is a standalone application too, we need to keep both versions.

nidalaa:
skip airbrake config in Rizzo **if** it is defined in app

robuye:
oh, sorry. Nop, can't do. Besides it inherits app api key.

nidalaa:
fair enough :+1:

robuye:
I want Rizzo on v5 some day :scream_cat:, opened an issue in airbrake.

PR #1469, opened 2016-02-11
Track flights and insurance submits as partner clicks

tkali:
Booking widgets that lead off site (as a result of form submit) should be considered as "Partner Clicks" not "Partner Searches".


WunderBart:
:+1:

PR #1499, opened 2016-03-23
Replace Sanitize with build-in helper

tkali:
Instead of using gem, we can rely on built-in `strip_tags` for HTML tags cleaning.


WunderBart:
:+1:

PR #1504, opened 2016-04-10
Tweak datepicker.js

WunderBart:
##### Tweaks/changes:
+ If user selects end date first, start date calendar will now open automatically
+ Other Picker will now open each time user starts the selection process
+ Setting invalid date will reopen the other Picker now
+ Setting same date is now available via `allowSameDate` option (it's helpful in situations when used as
f.e. date range in search filters)

**note**: _everything is backwards compatible and doesn't require taking any actions to upgrade._


thomasthesecond:
LGTM

PR #1505, opened 2016-04-11
Update npm packages

WunderBart:
+ Make versions less obscure
+ Move jasmine-jquery to node_modules instead of keeping it in the repo
+ Update Gruntfile.js


jcreamer898:
1 fail then :+1:

PR #1506, opened 2016-04-11
Fix randomly failing throttle.js spec

WunderBart:


jcreamer898:
nice

WunderBart:
Lesson learned: `tick()` doesn't work with `Date` until you actually `mockDate()` 😄

PR #1515, opened 2016-04-21
Use erb template for rizzo-next header

robuye:
When user registers we want to set proper source in Sailthru. If he
comes from Luna it should be `luna`, from Thorn Tree it should be
`community`, etc.

This can't be achieved with static generated HTML, not in the current
setup when file is not aware of the server role.

This will require additional manual step to update the template.
An `npm postinstall` instructions how to update the nav bar have been
added:

```
01:28:56 {ru-sailthru-source} ~/projects/rizzo$ npm install

> rizzo@0.1.2 postinstall /home/rob/projects/rizzo
> node rizzo-next-post-install.js

********************************************************
* To update global-header adjust the file:             *
*                                                      *
* app/views/layouts/_rizzo_next_global_header.erb.html *
*                                                      *
* to match the file:                                   *
*                                                      *
* node_modules/rizzo-next/dist/header.html             *
********************************************************
```

This is not ideal solution and works only in legazy rizzo applications,
but neither is the connection between rizzo-next and rizzo. Idealy we
would like to decouple both applications, make jenkins aware of app
identity and configure webpack to generate the file properly. However
right now there is no way to do this, and it's quite far road ahead of
us, so suboptimal solution has to do.

This is going to happen, `LP_SERVICE_ID` spreads like a disease all
over the place and we want to have it everywhere :smile_cat: 

/cc @WunderBart @jcreamer898

### DIFF
--- a/app/views/layouts/custom/_post_header.html.haml
+++ b/app/views/layouts/custom/_post_header.html.haml
@@ -4,7 +4,7 @@
 
 - if show_new_header?
   .content-wrapper.navigation-wrapper.rizzo-navigation.rizzo-next
-    = render :file => 'header'
+    = render 'layouts/custom/rizzo_next_global_header'
 - else
   .row--primary#js-primary-nav
     = render 'layouts/partials/snippets/primary_navigation_bar', search: search, user_nav: user_nav, include_js: include_js, nav_primary: nav_primary, responsive: responsive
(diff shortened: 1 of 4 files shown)


jcreamer898:
Ahh, if this is what you want, then I have an idea... :smirk:

robuye:
:smile_cat:  _is excited_

jcreamer898:
Ok, so, I'm struggling with it, will have to take some thinking... I was going to make rizzo next be a build tool type tying, but man, it's tough.

robuye:
It's not an easy problem with our current deployment pipeline. To make it work we will need to generate the file as part of assets building rather than commit it as static file in Rizzo.

To make it possible we will need to make Jenkins aware of an application identity first (`LP_SERVICE_ID`), it will be a part of migrating from current classic jobs to delivery pipeline (and even then we'd like to upgrade Jenkins to 2.0).

That makes me willing to take this shortcut until we have infrastructure in place to address the real problem.

PR #1516, opened 2016-04-22
Track pageView after lightbox content is ready

tkali:
This one handles page view tracking in appropriate place and ensures event is sent after lightbox renders its content with new datalayer.


WunderBart:
👍

Issue #1514, opened 2016-04-21
Map functionality is broken


lemmiestone:
If you hover your cursor over the map and then try enlarging it using the opposite-facing arrows, it doesn't work. However, If you don't hover over it, you can then click on the arrows to enlarge it - but you're then stuck with the enlarged map and can't reduce it back again.

laserchicken:
@lemmiestone Which map do you mean? Could you give me example url?

lemmiestone:
Sorry I meant to include one.  It's the maps on the POI pages.  http://www.lonelyplanet.com/usa/nashville/sights/architecture/ryman-auditorium

laserchicken:
it's fixed @lemmiestone

PR #1521, opened 2016-04-29
poi map resizing fix

laserchicken:
request: https://github.com/lonelyplanet/rizzo/issues/1514
ToggleActive is self-instantiating component, calling its constructor directly will register active class toggling callback second time - that way clicking `.js-toggle-active` will have no effect, since it will fire callbacks twice (setting back original state)


jcreamer898:
👍

PR #1524, opened 2016-05-24
update subnav (best in europe)

laserchicken:


jcreamer898:
Awesome, if you could just sort the links Alphabetically like they are on https://www.lonelyplanet.com/asia, then 👍

nicrocs:
👍

PR #1529, opened 2016-06-16
Improve adblock check with pixel bait

tkali:
This one adds pixel bait for adblockers. Checking for its visibility can detect some blockers which let `advertising.js` load.

Still, no match for µBlock (:heart:)


jcreamer898:
Seems legit b

PR #1531, opened 2016-06-24
Datepicker accepts min and max props

mdkalish:
Simple update for the datepicker to accept `min` and `max` properties.
```
= ui_component('forms/date_field', properties: { id: "text_id",
    label: "Choose Date",
    max: Date.today,
    min: "2010-01-01",
    name: "text_name",
})
```


WunderBart:
:shipit:

PR #1532, opened 2016-06-28
Bumping rizzo next

jcreamer898:


nicrocs:
looks like a problem with phantom.js install

PR #1533, opened 2016-06-29
Make the header load on shop

jcreamer898:
This change got missed when removing the split.


jcreamer898:
Don't love merging PRs w/o a 👍 but, gotta do what ya gotta do.

PR #1544, opened 2016-07-15
Bring back A9 and Criteo tags

tkali:
Due to problems with their implementation in Tealium (they introduced dependency on DFP; not added to Tealium yet).


jcreamer898:
👍

PR #1553, opened 2016-08-30
Fix thumb--tack's lp-blue icon option

WunderBart:
`lp-blue` color version of `thumb--tack` icon wasn't inlined vertically with the rest options.


jcreamer898:
👍

nicrocs:
👍

PR #1556, opened 2016-09-19
Prevent custom LightBox events bubbling

WunderBart:
Allowing events to bubble up caused overriding of every handler except last one (in most cases the last one attached to `#js-row--listener`).

This fix enables multiple listeners on multiple nodes (scopes).


nidalaa:
Works like a charm :shipit:

PR #1561, opened 2016-09-28
Mk click to copy colour

mdkalish:
@WunderBart, please review.


app/assets/javascripts/lib/styleguide/colours.js
```diff
@@ -77,4 +77,24 @@ define([ "jquery" ], function($) {
   if ($input.val()) {
     handleInput();
   }
+
+  // Click-to-copy functionality
+
+  var $colourBoxInput = $(".js-colour-box-input");
+
+  $(document).on("click", ".js-colour-box", function() {
```
WunderBart: I've noticed that clicking in FF makes page to jump (similar to when you click `#anchor`). I think we can put `event.preventDefault()` below to avoid that. Not sure if it helps, but it's worth trying ;)

nidalaa:
Wow, it looks and feels so nice! @WunderBart can we merge it :whip: :whip:?

mdkalish:
Right now I am working on the FF scrolling - it's `$.select()` what's causing it. In Chrome it doesn't scroll the page, unlike e.g. `focus()`, but in FF seemingly it does. Other than that the feature works correctly in FF.

app/assets/javascripts/lib/styleguide/colours.js
```diff
@@ -77,4 +77,24 @@ define([ "jquery" ], function($) {
...
+    $colourBoxInput
+      .val($this.data("clipboard-text"))
+      .select();
+    document.execCommand("copy");
+    $colourBoxInput.val("");
+
+    $this.addClass("is-copied");
+  });
+
+  $(document).on("mouseleave", ".js-colour-box", function() {
```
WunderBart: Can we please use `#js-row--content` as the listener here and [here](https://github.com/lonelyplanet/rizzo/pull/1561/files#diff-b4f1bcc7c07dc3e377be53d1de578affR85)? Can we also wrap it once and chain listeners?

WunderBart:
Would be super cool to have two more bugs fixed there:

1. Remove the `Colour list` section from the `Design palette`
2. Fix the `$nav-submenu-hover-border` block in the `UI Colours`

WunderBart:
Squash to single commit & :shipit:

PR #1574, opened 2016-12-13
Delete utag -> lp.analytics link

tkali:


robuye:
context: this line used to set `window.lp.analytics` to `undefined` in shop and break GA.

:+1:

PR #1576, opened 2016-12-19
Link legacy data layer objects to lp.dataLayer global

tkali:
This links legacy data objects used by Tealium and Omniture to `lp.dataLayer` global. 

`lp.dataLayer` will be prime data source for Google Tag Manager.


WunderBart:
👍

PR #1582, opened 2017-01-03
set ruby version to 2.0.0

robuye:
This is the version we have installed on production.


WunderBart:
quick, :shipit: :trollface:

PR #1594, opened 2017-03-27
Add Datepicker docs

WunderBart:
- Had to bump `autoprefixer-rails` to include the `appearance` prop which was not supported in current version. 
- Bumped `sass-rails` & fixed color maps warnings like this one:

  ```
  WARNING on line 28, column 10 of /Users/wunderbart/Projects/rizzo/app/assets/stylesheets/sass/mixins/_svg.sass:
  You probably don't mean to use the color value `peru' in interpolation here.
  It may end up represented as #cd853f, which will likely produce invalid CSS.
  Always quote color names when using them as strings (for example, "peru").
  If you really want to use the color value here, use `"" + $name'.
  ```
  👉  [how Sass compresses colors](https://github.com/sass/sass/blob/a76cb3f006141c5ca0c434b9f4f634d8bdbd04b3/doc-src/SASS_REFERENCE.md#colors)

- Added Datepicker documentation  
  
  screenshot 2017-03-27 20 50 14


jcreamer898:
Nice. 👍

PR #1599, opened 2017-04-21
Change default context for SelectGroupManager

WunderBart:
Because we sometimes want to use select in Lightbox and it's not under `.js-wrapper`.


nidalaa:
🐑

PR #1612, opened 2017-07-06
fix: Add the global header chunk

jcreamer898:


robuye:
I think this chunk is a bit too much? We don't exactly want apps to pull it all, right?

jcreamer898:
Yea, I'm going back to the drawing board with a new idea.

PR #1626, opened 2017-08-10
Prevent left nav's active item border from adding padding

WunderBart:
Before:

screenshot 2017-08-10 10 54 17

After: 

screenshot 2017-08-10 10 52 48


nidalaa:
Finally! That issue has been driving me crazy! :trollface:

PR #1628, opened 2017-09-01
fix: set default opt-in value

laserchicken:
related to: https://github.com/lonelyplanet/gonzo/pull/342
We should always set default opt-in value (`true`) to avoid passing `on/off` string. More details: https://stackoverflow.com/questions/12911787/html-checkbox-default-value. Thanks @robuye for investigating this.


robuye:
thanks @laserchicken !

is it possible to change email field to type `email`? I'm getting some errors for blank or invalid emails

laserchicken:
updated @robuye ^

Issue #1640, opened 2017-10-24
Italian Partner Site in footer results in error


chrissysievers:
**Short Description of Bug**
HTTPS links via the language menu at the bottom of the screen are bringing up net cert errors.

**Steps I should follow to reproduce**
go to https://www.lonelyplanet.com/ scroll to the language options at the bottom, choose Italiano. 

New info, this also happens to:
-India
-Italy
-Russia
-Brazil
-Czech

**What are the results you expect?**
NET::ERR_CERT_AUTHORITY_INVALID error

**Additional Info**
I spoke to Mike K about this, wondering if it had to do with the recent https transition. Apparently, this this link is not specifying https. 

![pasted image at 2017_10_24 04_13 pm](https://user-images.githubusercontent.com/20303852/31968992-9711d25a-b8d8-11e7-8698-feb5eabc5efb.png)

He suggested a front end see if there is a script redireting to ```https://{selected_value}```

Link to original [Bug Report](https://trello.com/c/YtHfQzek/405-https-links-via-the-language-menu-at-the-bottom-of-the-screen-are-bringing-up-net-cert-errors).

chrissysievers:
![](https://github.trello.services/images/mini-trello-icon.png) [HTTPS links via the language menu at the bottom of the screen are bringing up net cert errors.](https://trello.com/c/YtHfQzek/405-https-links-via-the-language-menu-at-the-bottom-of-the-screen-are-bringing-up-net-cert-errors)

laserchicken:
re "see if there is a script redireting to https://{selected_value}": yes, I can confirm this:  https://github.com/lonelyplanet/rizzo-next/blob/c554c38985de44b9afb828dd624dc7dcc58d1e83/src/components/footer/footer_component.js#L14

laserchicken:
re "wondering if it had to do with the recent https transition": yes, it's related (https://github.com/lonelyplanet/rizzo-next/commit/044e9464aa7bbd4c509e99ad7cfe31cb313db252)

laserchicken:
I can just change that redirect back to `http` and test - it should work I believe.

chrissysievers:
hey @laserchicken more info just came in on this one (I'll adjust the original issue above). Looks like more than Italian site is affected, and I didn't catch it..that's my mistake.

The Germans and French site have an https version that's working just fine, but these are also throwing errors:

-India - Seb to reach out to Indian
-Italy - International Licensing Team to manage
[...]

PR #1646, opened 2017-11-01
fix language selection

laserchicken:
request: https://github.com/lonelyplanet/rizzo/issues/1640

### DIFF
--- a/app/assets/javascripts/lib/core/language_nav.js
+++ b/app/assets/javascripts/lib/core/language_nav.js
@@ -1,0 +1,20 @@
+define([ "jquery" ], function($) {
+
+  "use strict";
+
+  function LanguageNav() {
+
+    $("#languageSelect").on("change", function(event) {
+      var url = "http://" + $(this).val();
+      window.location = url;
+      event.preventDefault();
+    });
+  }
+
+  $(document).ready(function() {
+    new LanguageNav;
+  });
+
+  return LanguageNav;
+
+});
--- a/app/assets/javascripts/lib/core/language_nav.js
+++ b/app/assets/javascripts/lib/core/language_nav.js
@@ -2,19 +2,26 @@ define([ "jquery" ], function($) {
 
   "use strict";
 
-  function LanguageNav() {
-
-    $("#languageSelect").on("change", function(event) {
-      var url = "http://" + $(this).val();
-      window.location = url;
-      event.preventDefault();
-    });
+  // @args = {}
+  // el: {string} selector for form
+  function LanguageNav(args) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/lib/core/language_nav.js
```diff
@@ -0,0 +1,20 @@
+define([ "jquery" ], function($) {
+
+  "use strict";
+
+  function LanguageNav() {
+
+    $("#languageSelect").on("change", function(event) {
```
WunderBart: Can we use `.js-language-select` here instead?

app/assets/javascripts/lib/core/language_nav.js
```diff
@@ -0,0 +1,20 @@
...
+      var url = "http://" + $(this).val();
```
WunderBart: Sure we want to force `http`?

laserchicken:
updated @WunderBart

app/assets/javascripts/lib/core/language_nav.js
```diff
@@ -0,0 +1,27 @@
...
+  };
+
+  LanguageNav.prototype.listen = function() {
+    this.$el.on("change", this._handleSelect);
+  };
+
+  LanguageNav.prototype._handleSelect = function(event) {
+    event.preventDefault();
+    var url = "http://" + $(this).val();
+    window.location = url;
```
WunderBart: I don't understand this one - do we redirect via `window.location` or [redirector form](https://github.com/lonelyplanet/rizzo/pull/1646/files#diff-0f67acf0410f857c2da18ffdc1b850a3R33)?

laserchicken: Sorry, I should remove all form related params - updated.
We are not using redirector form anymore, just simple dropdown select, and redirect with `window.location`, exactly the same as it's implemented in [rizzo-next](https://github.com/lonelyplanet/rizzo-next/blob/master/src/components/footer/footer_component.js#L16)

laserchicken: Redirector endpoint (after we switched to https [here](https://github.com/lonelyplanet/rizzo/commit/a51be21a0ad992059248280b54a1e76c10f40e65#diff-0f67acf0410f857c2da18ffdc1b850a3R43)) started forcing https for all redirects and some of those external pages do not support it. We need to make it work and the only option is  to change it back to http for now (and we can not use redirector endpoint anymore).

laserchicken: Does it make sense?

PR #1672, opened 2018-08-27
Set default dates from the 'value' html attr

mdkalish:
Solves the bug that dates are not picked up from the 'value' and prepopulated at page reload.
Example: https://www.lonelyplanet.com/thorntree/travel_companions

### DIFF
--- a/app/assets/javascripts/lib/components/availability_search.js
+++ b/app/assets/javascripts/lib/components/availability_search.js
@@ -92,7 +92,7 @@ define([
   };
 
   AvailabilitySearch.prototype._getSearchData = function() {
-    if (!this.$startDate.val()) {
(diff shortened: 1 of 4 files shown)


WunderBart:
Could you check those failing specs, please?

WunderBart:
1. Shouldn't the pre-populated input value be used as a [default from/to date](https://github.com/lonelyplanet/rizzo/blob/master/app/assets/javascripts/lib/components/availability_search.js#L95-L97)?
2. Please don't use dynamic date in specs - it's always caused troubles,

mdkalish:
[Airbrake fusspot: ArgumentError: invalid date](https://trello.com/c/lLT2X7Tk/1565-airbrake-fusspot-argumenterror-invalid-date)

PR #1678, opened 2018-11-12
[ADS-141] include user-status ssi tag

johnwquarles:
Puts `<!--#include virtual="/global-ui/user-status" -->` (that's the extent of the PR) into `_head.html.haml`.

This will cover the major `rizzo` layouts (`minimal`, `responsive`, and `styleguide`).

End goal is to have this tag included such that user authentication details are available for tracking purposes.

I'm going to start by utilizing this data in `gonzo` and move forward from there; I think we're going to focus on these legacy apps in the next sprint. This will facilitate that effort.

### DIFF
--- a/app/views/layouts/custom/_head.html.haml
+++ b/app/views/layouts/custom/_head.html.haml
@@ -24,6 +24,8 @@
 
 = csrf_meta_tags
 
+<!--#include virtual="/global-ui/user-status" -->
+
 - if include_js
   = render 'layouts/partials/snippets/head_js', third_party: third_party
 


johnwquarles:
oh forgot about this one

### Please Do Not Merge

johnwquarles:
I've decided to add this markup in to individual codebases.

Reasoning:

> Thinking about this and other (older) codebases that use rizzo:

> Ultimately the ssi markup could be added via rizzo, meaning that we’d only have to bump the rizzo sha in each codebase where we need the markup to appear.

> But, I think it’s more readable/immediately understandable for devs to have that markup present in the host codebase itself. And it’s only one line.

> Ultimately, that approach will still contain a dependency on rizzo in that all the analytics content is injected, as one unit, into a block provided by rizzo's layout (- content_for(:lp_analytics)`)

> Nonetheless, if rizzo were ever scrapped in these codebases (which I realize will never happen, but on principle), we could just put that block somewhere else in the host codebase layout, which would need to be added at such time as rizzo were removed anyhow.

> Also, if we/I were serious about leveraging rizzo to share any of this analytics code, I should have used the block to inject only the actual datalayer-initialized payload, not it and the initialization scripts that surround it.

> For these reasons, I favor adding the markup independently in each codebase, at the top of the content_for(:lp_analytics) block.

> Maybe if rizzo were still our standard layout across all LP sites, it’d be worth leveraging it to share this code. As-is, though, I feel like it makes sense to move away from it as much as possible-- to detangle it from the code that is still in use, and make everything more straightforward for devs, if more verbose.

PR #1699, opened 2020-11-05
Footer updates and documentation

vacas5:
Updated documentation to not reference gonzo because the newsletter route doesn't support GET anymore. Can simply
test changes locally.

Changed footer links to be more like our current ones in DFE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Rizzo
 
-Rizzo is the UI layer for lonelyplanet.com. Rizzo also serves LP's header and footer, assets and Style Guide.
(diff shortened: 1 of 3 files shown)


laserchicken (approved):

jcreamer898:
Haha, where is Rizzo even still used?

katelynp43:
@jcreamer898 haha, you'd be surprised of all those legacy systems that are still lingering... haha Never ending retirement. 😉 🤣

laserchicken:
👋 haha

